PR #1, opened 2024-02-29
Initial boilerplate

asgrim:
Some assumptions/opinions made for the development of the tool are:

 * Minimum PHP version 8.1, given that is the only PHP version in security support (or greater)
 * PHPUnit for unit testing; fairly strongly opinionated on this, but again, is a preference.
 * Psalm for static analysis; not strongly opinionated here, but I have more experience working with Psalm, so this is my preference
 * `doctrine/coding-standard` for CS; I feel this CS is comprehensive and fixes most code in a predictable way, and I am most familiar with `phpcs` (over say, php-cs-fixer).
 * I have used `laminas/workflow-continuous-integration` to run the CI workflow, as it is ludicrously easy to get up and running. Should we need to expand out of these realms in the future for static analysis/CS/linting/testing etc., we can of course


composer.json
```diff
@@ -0,0 +1,39 @@
...
+        "psr-4": {
+            "Php\\Pie\\": "src/"
+        }
+    },
+    "autoload-dev": {
+        "psr-4": {
+            "Php\\PieUnitTest\\": "test/unit/"
+        }
+    },
+    "authors": [
```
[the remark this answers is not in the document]
asgrim: Done

test/unit/Command/DownloadCommandTest.php
```diff
@@ -0,0 +1,28 @@
...
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+
+#[CoversClass(DownloadCommand::class)]
+class DownloadCommandTest extends TestCase
+{
+    public function testDownloadCommand(): void
+    {
+        $input  = $this->createMock(InputInterface::class);
+        $output = $this->createMock(OutputInterface::class);
```
Seldaek, line 19: Just FYI as you may not be aware of it, there's a tester class for commands integration testing, see e.g.

https://github.com/composer/composer/blob/main/tests/Composer/Test/TestCase.php#L201-L211 (the ApplicationTester class is part of symfony/console tho)

asgrim: Thanks - I didn't know about this. I've switched to using `CommandTester` in a follow-up branch to this, so will be addressed then :+1:

composer.json
```diff
@@ -0,0 +1,43 @@
...
+            "name": "James Titcumb",
+            "email": "asgrim@php.net"
+        },
+        {
+            "name": "PIE contributors",
+            "homepage": "https://github.com/php/pie/graphs/contributors"
+        }
+    ],
+    "require": {
+        "php": "8.1.*||8.2.*||8.3.*",
```
asgrim, line 28: To be discussed: ThePHPF/pie-design#7

test/unit/Command/DownloadCommandTest.php
theofidry, line 19: I wonder why you're using a mock here? Wouldn't be passing `StringInput('')` and `NullOutput` be sufficient?

composer.json
asgrim: Resolution as per https://github.com/ThePHPF/pie-design/issues/7#issuecomment-2003177751

> I think therefore, this discussion can probably be closed. As per PR #1 the tool will initially need PHP 8.1 to be run, but will still be able to install older PHP versions using `--with-php-config`, as mentioned.

composer.json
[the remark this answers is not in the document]
Tomyh99095: > Add an author:
> 
> ```json
>         {
>             "name": "PIE contributors",
>             "homepage": "https://github.com/php/pie/graphs/contributors"
>         }
> ```

Issue #13, opened 2024-05-27
Add support for `dev-main` and `dev-main#sha` type version requests


asgrim:
`\Php\PieIntegrationTest\Command\DownloadCommandTest::validVersionsList` has commented out `dev-main` and `dev-main#<commit-sha>` tests, as they aren't currently working. In theory it should be possible to get this working, but need to check into why.

PR #19, opened 2024-07-10
Build Command implementation

### LINKED ISSUE #3: Builder command & related components

Implement the "Builder" command, and its related components:

 * Linux builder
 * Windows builder (no-op, initially)

Definition of done:

 - A Linux user is able to download and build a downloaded package using a command like `bin/pie build ext-foo:^1.2`

asgrim:
Fixes #3 

I'm going to merge this one fairly soon, as it isn't a huge add. Assuming there is no show-stoppers, I'll probably merge end of this week, or beginning of next, just in case anyone wants to give any further feedback.

# Are you an end user who would like to try this out?

Please note that this is just an initial implementation, and will **only download and build** an extension to a temporary path on your machine; it will _not install anything_ just yet. That is coming soon!

**If you understand, and you'd still like to help test this...**

Primarily the best way to provide feedback here as an end user of PIE is to check the `build` command works on your machine! You can do that by following these steps (please adjust for your platform differences of course!)

 * Ensure you are using **PHP 8.3** (this is because my test extension `asgrim/example-pie-extension` is intentionally only compatible with PHP 8.3)
 * `git clone -b download-implementation https://github.com/asgrim/pie.git`
 * `composer install`
 * `bin/pie build asgrim/example-pie-extension`
 * Let me know if you encounter any errors or something unexpected!

You should see something like:

```
$ bin/pie build asgrim/example-pie-extension
You are running PHP 8.3.8
Target PHP installation: 8.3.8 nts, on Linux/OSX/etc x86_64 (from /usr/bin/php8.3)
phpize complete.
Configure complete.
Build complete.
```

# Are you an extension maintainer who would like to try this out?

It's not strictly necessary to support this just yet, but if you're keen, and please note, things may change before we finally release PIE... so you do this at your own risk. Note: at the moment, you will need to make a new release - see ThePHPF/pie-design#17

 * Add and commit a `composer.json` to your repo - example here https://github.com/asgrim/example-pie-extension/blob/main/composer.json
 * Make a **release** (`alpha` release is fine, e.g. `1.2.3-alpha.`) - note, "branches" or "commits" are not yet supported - see #13 
 * Then, add it to Packagist here: https://packagist.org/packages/submit (you put the GH URL in, it should detect the `composer.json` automatically and understand that it is a `php-ext` or `php-ext-zend`, it should appear in https://packagist.org/extensions once added

More details, for example for help on `composer.json`, can be read in https://github.com/ThePHPF/pie-design?tab=readme-ov-file#extension-maintainer-register-a-pie-package


src/Building/UnixBuild.php
```diff
@@ -0,0 +1,69 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Php\Pie\Building;
```
theofidry, line 5: `Building` looks a bit weird, given that the initial implementation seems to be more zend/Symfony-esque, guess you didn't like having `Build` there?

asgrim: `Building` is just the domain concern here, no strong opinions on frameworks here, it just is what it is :shrug:

src/Building/UnixBuild.php
```diff
@@ -0,0 +1,69 @@
...
+    /** {@inheritDoc} */
+    public function __invoke(
+        DownloadedPackage $downloadedPackage,
+        array $configureOptions,
+        OutputInterface $output,
+    ): void {
+        $this->phpize($downloadedPackage);
+        $output->writeln('<info>phpize complete</info>.');
+
+        $this->configure($downloadedPackage, $configureOptions);
```
asgrim, line 28: Need to add `--with-php-config=...` if specificed in CLI

asgrim: Done

PR #25, opened 2024-08-02
Check if a given PHP Binary is actually valid

### LINKED ISSUE #12: Verify that the PhpBinaryPath created is really a PHP

asgrim:
Fixes #12

### DIFF
--- a/src/Platform/TargetPhp/PhpBinaryPath.php
+++ b/src/Platform/TargetPhp/PhpBinaryPath.php
@@ -41,22 +41,27 @@ class PhpBinaryPath
         public readonly string $phpBinaryPath,
         private readonly string|null $phpConfigPath,
     ) {
-        if (! file_exists($this->phpBinaryPath)) {
-            throw Exception\InvalidPhpBinaryPath::fromNonExistentPhpBinary($this->phpBinaryPath);
+    }
+
+    /** @param non-empty-string $phpBinaryPath */
+    private static function assertValidLookingPhpBinary(string $phpBinaryPath): void
+    {
+        if (! file_exists($phpBinaryPath)) {
+            throw Exception\InvalidPhpBinaryPath::fromNonExistentPhpBinary($phpBinaryPath);
         }
 
-        if (! is_executable($this->phpBinaryPath)) {
-            throw Exception\InvalidPhpBinaryPath::fromNonExecutablePhpBinary($this->phpBinaryPath);
+        if (! is_executable($phpBinaryPath)) {
+            throw Exception\InvalidPhpBinaryPath::fromNonExecutablePhpBinary($phpBinaryPath);
         }
 
         // This is somewhat of a rudimentary check that the target PHP really is a PHP instance; not sure why you
         // WOULDN'T want to use a real PHP, but this should stop obvious hiccups at least (rather than for security)
-        $testOutput = trim((new Process([$this->phpBinaryPath, '-r', 'echo "PHP";']))
+        $testOutput = trim((new Process([$phpBinaryPath, '-r', 'echo "PHP";']))
             ->mustRun()
             ->getOutput());
 
         if ($testOutput !== 'PHP') {
-            throw Exception\InvalidPhpBinaryPath::fromInvalidPhpBinary($this->phpBinaryPath);
+            throw Exception\InvalidPhpBinaryPath::fromInvalidPhpBinary($phpBinaryPath);
         }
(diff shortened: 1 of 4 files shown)


src/Platform/TargetPhp/PhpBinaryPath.php
```diff
@@ -31,7 +33,23 @@ private function __construct(
         public readonly string $phpBinaryPath,
         private readonly string|null $phpConfigPath,
     ) {
-        // @todo https://github.com/php/pie/issues/12 - we could verify that the given $phpBinaryPath really is a PHP install
+        if (! file_exists($this->phpBinaryPath)) {
```
theofidry, line 36: maybe for testing purposes you may want to be able to create an instance without caring about whether it is valid or not, or for performance reasons in a context where you're sure it is valid. Might be worth having this logic in a static factory? (Could be just `PhpBinaryPath::create()`).

asgrim: Yep that seems reasonable, I'll adjust, thank you

ghostwriter:
You are using the wrong directory separator on windows, possibly causing is_executable to “always” return false.

asgrim:
> You are using the wrong directory separator on windows, possibly causing is_executable to “always” return false.

Hmm, looking at this test:

https://github.com/php/php-src/blob/589cfbb24fde06836df9dbb50ee0093a69c6eeba/ext/standard/tests/file/windows_acls/bug44859_3.phpt#L13-L31

The commented-out `tiny.bat` suggests that the `.bat` workaround that I tried won't work. I think it would have to be a true executable on Windows. That suggests the doc:

> On Windows, a file is considered executable, if it is a properly executable file as reported by the Win API GetBinaryType();

is correct, but:

> for BC reasons, files with a .bat or .cmd extension are also considered executable.

is outdated.

ghostwriter:
I’m talking about the path you hardcoded in the test.

image

`/ for Unix-based systems (Linux, macOS).`

`\ for Windows.`

use `DIRECTORY_SEPARATOR` to automatically use the correct one for the platform at runtime.

ghostwriter:
The path: `"z:\phpf\pie\test\unit\Platform\TargetPhp/../../assets/fake-php-windows.bat”`

using both `\` and `/` in the path.


image

asgrim:
> The path: `"z:\phpf\pie\test\unit\Platform\TargetPhp/../../assets/fake-php-windows.bat”`
> 
> using both `\` and `/` in the path.
> 
> 
> image
> 

Yes, I understand what you're saying, but I tried fixing the slashes since my earlier stream, and that is still not the issue. I explored the code some more, and the .bat/.cmd thing does not appear hold true.

ghostwriter:
I see, "normalizing" the path was just the **first** step.

--- 

> … the .bat/.cmd thing does not appear hold true.

you are correct, it does **not** appear hold `true` in this case, but the code does reflect what is documented.

https://github.com/php/php-src/blob/PHP-8.3.11/win32/ioutil.c#L924-L940

--- 

My understanding was, and please correct me If i’m wrong, even if we change the file extension, on windows, if the file’s internal format is **NOT** recognized as executable, `is_executable()` will return `FALSE` (except for file paths with `*.bat` and `*.cmd` extension).

Executability is determined by:

**Windows** returns `TRUE`: 

- If the given file path exists and has a `*.exe` or `*.com` extension, with the [PE format](https://en.wikipedia.org/wiki/Portable_Executable) and internal structure that is recognized as executable using`GetBinaryTypeW`.

**or** 

- If the given file path exists and has a `*.bat` or `*.cmd` extension

**Unix-like Systems** returns `TRUE`: 

- If the given file path exists and the file has the executable bit set.

---

I would not say [outdated](https://github.com/php/php-src/blob/PHP-8.3.11/win32/ioutil.c#L924-L940) but there might be an undocumented behavior/s when we [set the file type and permissions](https://github.com/php/php-src/blob/PHP-8.3.11/win32/ioutil.c#L941-L957) or when [checking executable permission](https://github.com/php/php-src/blob/PHP-8.3.11/win32/ioutil.c#L671-L674) or on files with [empty DACLs](https://github.com/php/php-src/blob/PHP-8.3.11/win32/ioutil.c#L886). 

(empty DACL: no permissions explicitly granted or denied to anyone, effectively making the file inaccessible to all users, including the file's owner.)

---

P.S. Thanks for being patient and indulging my curiosity.

PR #33, opened 2024-09-10
Only use sudo if needed

### LINKED ISSUE #29: Don't require `sudo` if we can already write to target(s)
At the moment, `pie install` will ask for `sudo` regardless, even if we can already write to the target path(s). 
We should improve this to check if we actually NEED to even ask.

asgrim:
Fixes #29

### DIFF
--- a/test/integration/Command/InstallCommandTest.php
+++ b/test/integration/Command/InstallCommandTest.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Php\PieIntegrationTest\Command;
 
+use Composer\Util\Platform;
 use Php\Pie\Command\InstallCommand;
 use Php\Pie\Container;
 use PHPUnit\Framework\Attributes\CoversClass;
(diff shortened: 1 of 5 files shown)


test/integration/Command/InstallCommandTest.php
```diff
@@ -62,6 +64,44 @@ public function testInstallCommandWillInstallCompatibleExtension(): void
...
+            array_unshift($rmCommand, 'sudo');
+        }
+
+        (new Process($rmCommand))->mustRun();
+    }
+
+    public function testInstallCommandWillInstallCompatibleExtensionWindows(): void
+    {
+        if (PHP_VERSION_ID < 80300 || PHP_VERSION_ID >= 80400) {
+            self::markTestSkipped('This test can only run on PHP 8.3 - you are running ' . PHP_VERSION);
```
derickr, line 80: Why is this PHP 8.3 only?

asgrim: Testing is tricky unless you have multiple versions of PHP available (e.g. using deb.sury.org, or you compile several different versions of PHP); therefore to reduce reliance on that particular complexity, this test only runs in PHP 8.3. Can certainly be worked on, but it will make CI much more complex :sweat_smile:

Issue #20, opened 2024-07-11
Detect and automatically set up `php.ini` entry for installation


asgrim:
Typically, the approach for adding `extension=blah.so` or `zend_extension=blah.so` to the PHP INI would be:

 * add a new file with the line in the directory indicated by `Scan for additional .ini files in:` using the priority defined in `composer.json` (e.g. `20-myext.ini`)
 * if the `Scan for additional .ini files in:` is empty, add the line to the file indicated by `Loaded Configuration File:` entry
 * if that is also empty, tell the user what they need to do

However, when using setups like the `deb.sury.org` PHP installer, this approach would ONLY enable the ext for a single SAPI (the CLI SAPI). This would be confusing as the extension would not be enabled for the Apache SAPI, so consumers may be confused by the fact it does not have the extension loaded!

Investigate the various common approaches to how `php.ini` are used, and we may need to add special handling for each "custom" approach, and then fall back to the expected typical approach above.

Ayesh:
PHP in Debian and derivatives come with a `php(en|dis)mod` script, that I think solves this problem fairly well.  It accepts a SAPI name (_all SAPIs by default), and creates a in formats such as `20-mbstring.ini` that only has `extension=mbstring.so`. I like that it does not modify the main `php.ini` file, but instead adds/removes these "loader" files in all/selected SAPI-specific `conf.d` folders.

I suppose we can use `php_ini_loaded_file()`/` PHP_INI_SCAN_DIR`/`php_ini_scanned_files()` to try and guess if we are in a Debian/RHEL-like environment, and if we detect a `conf.d` in one of the scanned file paths, to add a "loader" ini file for the extension.

On Windows, I don't think `PHP_INI_SCAN_DIR` is actively used, so in all other situations, we can modify the ini file.

ruudk:
When implementing this, please allow for a CLI option to disable it.

Currently, we use PECL to install extensions like xdebug. For reasons, we want xdebug to be installed, but not enabled. We always have to do post processing to remove the auto enabled extension from php.ini or conf.d/99-xdebug.ini.

So for PIE, it would be super nice to have `--no-enable` or something.

TimWolla:
> Investigate the various common approaches to how `php.ini` are used, and we may need to add special handling for each "custom" approach, and then fall back to the expected typical approach above.

Within the `php` Docker Official Image it should be [`docker-php-ext-enable`](https://github.com/docker-library/php/blob/master/docker-php-ext-enable).

PR #34, opened 2024-09-10
Added initial documentation for maintainers

### LINKED ISSUE #8: Documentation for extension maintainers

Produce documentation for extension maintainers on how to make their extension available with PIE.

asgrim:
Fixes #8

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,53 +1,48 @@
 # 🥧 PIE (PHP Installer for Extensions)
 
-You will need PHP 8.1 or newer to run PIE, but PIE can install an extension to any installed PHP version.
+You will need PHP 8.1 or newer to run PIE, but PIE can install an extension to
+any installed PHP version.
 
-## Installing
+If you are an extension maintainer wanting to add PIE support to your extension,
+please read [extension-maintainers](./docs/extension-maintainers.md).
 
-## Usage
+## Installing an extension using PIE
 
-You can download an extension ready to be built or installed using the `download` command. For example, to download the
-`example_pie_extension` extension, you would run:
+You can install an extension using the `install` command. For example, to
+install the `example_pie_extension` extension, you would run:
 
 ```shell
-$ bin/pie download asgrim/example-pie-extension
-You are running PHP 8.3.7
-Target PHP installation: 8.3.7 nts, on Linux/OSX/etc x86_64 (from /usr/bin/php8.3)
+$ bin/pie install asgrim/example-pie-extension
+This command may need elevated privileges, and may prompt you for your password.
+You are running PHP 8.3.10
+Target PHP installation: 8.3.10 nts, on Linux/OSX/etc x86_64 (from /usr/bin/php8.3)
 Found package: asgrim/example-pie-extension:1.0.1 which provides ext-example_pie_extension
-Extracted asgrim/example-pie-extension:1.0.1 source to: /tmp/pie_downloader_6645f07a28bec9.66045489/asgrim-example-pie-extension-769f906
-$ 
+phpize complete.
+Configure complete.
+Build complete: /tmp/pie_downloader_66e0b1de73cdb6.04069773/asgrim-example-pie-extension-769f906/modules/example_pie_extension.so
+Install complete: /usr/lib/php/20230831/example_pie_extension.so
+You must now add "extension=example_pie_extension" to your php.ini
+$
 ```
 
-If you are trying to install an extension for a different version of PHP, you may specify this on non-Windows systems
-with the `--with-php-config` option like:
+### Using PIE to install an extension for a different PHP version
+
+If you are trying to install an extension for a different version of PHP, you
+may specify this on non-Windows systems with the `--with-php-config` option:
 
 ```shell
-bin/pie download --with-php-config=/usr/bin/php-config7.2 my/extension
+bin/pie install --with-php-config=/usr/bin/php-config7.2 my/extension
 ```
 
-On all platforms, you may provide a path to the `php` executable itself using the `--with-php-path` option. This is an
-example on Windows where PHP 8.1 is used to run PIE, but we want to download the extension for PHP 8.3:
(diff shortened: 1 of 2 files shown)


docs/extension-maintainers.md
```diff
@@ -0,0 +1,128 @@
...
+The first step to adding PIE support is adding a `composer.json` to your
+extension repository. Most of the typical fields are the same, with a few
+notable exceptions:
+
+ * The `type` MUST be either `php-ext` for a PHP Module (this will be most
+   extensions), or `php-ext-zend` for a Zend Extension.
+ * An additional `php-ext` section MAY exist.
+ * The Composer package name (i.e. the top level `name` field) for a PIE
+   extension cannot share the same Composer package name as a regular PHP
+   package, even if they have different `type` fields.
```
alcaeus, line 17: Maybe it's worth mentioning that the package name should follow the usual convention for Composer packages (i.e. `<vendor>/<package>` instead of the "usual" `ext-<extension>`) in addition to being unique across the Composer ecosystem.

docs/extension-maintainers.md
```diff
@@ -0,0 +1,128 @@
...
+{
+    "name": "myvendor/myextension"
+}
+```
+
+The extension name would be derived as `myextension`. The `myvendor/` vendor
+prefix is removed.
+
+The `extension-name` MAY be prefixed with `ext-` as is a convention in Composer,
+but this is optional.
```
alcaeus, line 40: I don't remember all of the context behind this decision, but as a best practice for new authors I think it's better to suggest people don't prefix their extension name, as `ext-` is only used to distinguish extensions from packages and libraries in an otherwise ambiguous name field. Since we're in the context of a PHP extension here, the `ext-` prefix is implied and should be omitted from the name.

Consider this:
```suggestion
The `extension-name` SHOULD NOT be prefixed with `ext-` as is a convention in Composer.
```

Instead of suggesting that users can use the prefix, this wording suggests that authors omit it, without saying that it's forbidden. I understand that leaves a grey area ("is it supported?"), but I'd much rather avoid a useless prefix.

docs/extension-maintainers.md
```diff
@@ -0,0 +1,128 @@
...
+
+Note that it is not possible for end users of PIE to specify configuration
+options that have not been defined in your extension's `configure-options`
+definition. Using the same example above `composer.json`, invoking PIE with
+an invalid option, such as `pie install myvendor/myext --something-else` will
+result in an error `The "--something-else" option does not exist.`.
+
+If an end user does not specify a flag defined in the `configure-options`
+definition, it will simply not be passed to `./configure`. There is no way to
+specify a default value in the `configure-options` definition.
```
alcaeus, line 120: Thank you for this entire chapter - that cleans up most of the questions that came up while I was adding PIE support for the MongoDB extension.

docs/extension-maintainers.md
```diff
@@ -0,0 +1,128 @@
...
+
+## Submit the extension to Packagist
```
alcaeus, line 122: Could you add a note that package authors also need to tag a release for the extension to be installable until support for building from commit references is added? Otherwise, this may confuse users that are used to not needing a tag to test a package.

alcaeus (commented):
Thank you for adding the maintainer docs. I have a couple of suggestions for clarifications, but this is very helpful for extension authors!

README.md
```diff
@@ -1,53 +1,48 @@
...
-You can download an extension ready to be built or installed using the `download` command. For example, to download the
-`example_pie_extension` extension, you would run:
+You can install an extension using the `install` command. For example, to
+install the `example_pie_extension` extension, you would run:
 
 ```shell
-$ bin/pie download asgrim/example-pie-extension
-You are running PHP 8.3.7
-Target PHP installation: 8.3.7 nts, on Linux/OSX/etc x86_64 (from /usr/bin/php8.3)
+$ bin/pie install asgrim/example-pie-extension
```
derickr, line 15: Perhaps it'd be wise to use `example/` instead of `asgrim/`?

README.md
```diff
@@ -1,53 +1,48 @@
...
+You are running PHP 8.3.10
+Target PHP installation: 8.3.10 nts, on Linux/OSX/etc x86_64 (from /usr/bin/php8.3)
 Found package: asgrim/example-pie-extension:1.0.1 which provides ext-example_pie_extension
-Extracted asgrim/example-pie-extension:1.0.1 source to: /tmp/pie_downloader_6645f07a28bec9.66045489/asgrim-example-pie-extension-769f906
-$ 
+phpize complete.
+Configure complete.
+Build complete: /tmp/pie_downloader_66e0b1de73cdb6.04069773/asgrim-example-pie-extension-769f906/modules/example_pie_extension.so
+Install complete: /usr/lib/php/20230831/example_pie_extension.so
+You must now add "extension=example_pie_extension" to your php.ini
```
derickr, line 24: I thought the idea was to drop configuration files in the `{config_dir}/conf.d/` directory automatically?

docs/extension-maintainers.md
derickr: AFAIK, you can already do this, or at least from `dev-master`: 
```
pie install xdebug/xdebug:dev-master
```

asgrim: Since #13 was fixed, it should be possible as mentioned. I'll check Mongo shortly, and if it's not working try to figure out why separately

README.md
asgrim: I was copy/pasta from my terminal to ensure we had "real" output, but yeah, I'll replace this :+1:

README.md
asgrim: Yes, there is an issue #20 to be addressed to add this functionality :+1:

docs/extension-maintainers.md
```diff
@@ -0,0 +1,130 @@
+# PIE for Extension Maintainers
```
asgrim, line 1: Need to add a section about Windows and how DLLs need to be published.

PR #37, opened 2024-09-16
Support numeric-branch -dev branches

### LINKED ISSUE #36: Support <numeric-branch>-dev branches
The [mongodb](https://github.com/mongodb/mongo-php-driver) extension has **numeric** branch names, which are inversed with version requirements (i.e. `<numeric-branch>-dev` vs `dev-<named-branch`).

Therefore, we should update the `\Php\Pie\DependencyResolver\ResolveDependencyWithComposer::factoryRepositorySet` to check for numeric branches too.

asgrim:
Fixes #36


src/DependencyResolver/DetermineMinimumStability.php
```diff
@@ -0,0 +1,64 @@
...
+            ],
+        );
+    }
+
+    /** @return self::STABILITY_* */
+    public static function fromRequestedVersion(string|null $requestedVersion): string
+    {
+        /** Stability options from {@see https://getcomposer.org/doc/04-schema.md#minimum-stability} */
+        if ($requestedVersion !== null) {
+            if (preg_match('#@(dev|alpha|beta|RC|stable)$#', $requestedVersion, $matches)) {
```
Seldaek, line 49: This should be case insensitive IMO

```suggestion
            if (preg_match('#@(dev|alpha|beta|RC|stable)$#i', $requestedVersion, $matches)) {
```

But also maybe you should just use https://github.com/composer/semver/blob/main/src/VersionParser.php#L52C28-L52C42

asgrim: Yes! Nice one, thank you, updated :100:

PR #46, opened 2024-09-23
Check ZTS and NTS flags when resolving package

### LINKED ISSUE #35: Add check for ZTS compatibility
The PIE design doc specifies:

> * The package will be checked to see if it is compatible with ZTS mode. If using
>   a `--with-php-config` with ZTS
>   enabled, but the extension specifies `"support-zts": false` in
>   its `composer.json` manifest, the installation will
>   be halted with an error explanation.

This check doesn't currently happen, but should be added.

asgrim:
Fixes #35

### DIFF
--- a/src/DependencyResolver/Package.php
+++ b/src/DependencyResolver/Package.php
@@ -29,6 +29,8 @@ final class Package
         public readonly array $configureOptions,
         public readonly string|null $notificationUrl,
         public readonly string $notificationVersion,
+        public readonly bool $supportZts,
+        public readonly bool $supportNts,
     ) {
     }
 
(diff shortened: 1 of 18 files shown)


src/DependencyResolver/Package.php
```diff
@@ -43,6 +45,14 @@ public static function fromComposerCompletePackage(CompletePackageInterface $com
             )
             : [];
 
+        $supportZts = $phpExtOptions !== null && array_key_exists('support-zts', $phpExtOptions)
+            ? $phpExtOptions['support-zts']
+            : true;
+
+        $supportNts = $phpExtOptions !== null && array_key_exists('support-nts', $phpExtOptions)
```
asgrim, line 52: Need upstream composer/composer#12127 and to update to use this merged revision

asgrim: Merged and updated here.

PR #76, opened 2024-10-24
Parallel jobs with make

### LINKED ISSUE #73: Use parallel build
For now make is run using a single processor

On linux, number of available processor can be retrieved using

```
$ getconf _NPROCESSORS_ONLN
64
```

Running `make -j64` will be obviously faster, especially for big sources (grpc, couchbase, ddtrace...)

The above value can be used, or perhaps a new build option

asgrim:
Fixes #73 

PIE will try three approaches to determine the number of cores (`nproc --all`, `getconf __NPROCESSORS_ONLN`, `sysctl -n hw.ncpu`), or you can override with `-jN` or `--make-parallel-jobs=N` when running with PIE.

Doesn't cover Windows; or maybe some uncommon systems without those above commands. Suggestions welcome.

### DIFF
--- a/src/Util/NumberOfCores.php
+++ b/src/Util/NumberOfCores.php
@@ -1,0 +1,69 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Php\Pie\Util;
+
+use Symfony\Component\Process\Process;
+
+final class NumberOfCores
+{
+    /** @psalm-suppress UnusedConstructor */
+    private function __construct()
+    {
+    }
+
+    /**
+     * @param list<string> $params
+     *
+     * @return positive-int
+     */
+    private static function tryCommand(string $command, array $params): int|null
+    {
+        $whichProcess = new Process(['which', $command]);
+        if ($whichProcess->run() === 0) {
+            $commandPath = trim($whichProcess->getOutput());
+
+            $commandProcess = new Process([$commandPath, ...$params]);
+            if ($commandProcess->run() === 0) {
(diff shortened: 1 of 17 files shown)


remicollet:
LGTM :)

remicollet:
@asgrim see https://github.com/asgrim/pie/pull/2

I really think (especially in case of long build) user should be able to follow what happen

asgrim:
> @asgrim see [asgrim#2](https://github.com/asgrim/pie/pull/2)
> 
> I really think (especially in case of long build) user should be able to follow what happen

Thanks - I made a new ticket #77 to handle separately :+1: good suggestion, thank you

src/Util/NumberOfCores.php
```diff
@@ -0,0 +1,72 @@
...
+        return null;
+    }
+
+    /**
+     * Try a few different ways of determining number of CPUs. If we can't guess
+     * how many CPUs someone has, then default to 1 for safety.
+     *
+     * @return positive-int
+     */
+    public static function determine(): int
```
theofidry, line 52: I recommend https://github.com/theofidry/cpu-core-counter, its the code used in most tools which have parallelisation

llaville: Some great packages already used it : see https://packagist.org/packages/fidry/cpu-core-counter/dependents?order_by=downloads
And I hope (me too) to adapt some of my packages to add a parallelization feature

Issue #31, opened 2024-09-07
Audit dependencies


asgrim:
We are currently using some dependencies which MAY be problematic for maintenance.

Whilst not a big issue, if we want to support newer versions of PHP (e.g. 8.4 at the time of writing, is unreleased), we have to ensure we are compatible with the newer dependencies. For example, the constraint we have for `illuminate/container` is `^10.47`, which would not allow version 11 (which I believe has fixed the PHP 8.4 issues). This is somewhat mitigated by our root `php` constraint being explicitly `8.1.*||8.2.*||8.3.*` (i.e., we do not yet support PHP 8.4), and this issue only surfaces when `--ignore-platform-req=php` is used (e.g. for testing). The more third-party dependencies we have, the more we are constrained by upstream support. We should evaluate the dependencies we are using, and ensure it makes sense to continue using them or not.

PSL - https://packagist.org/packages/azjezz/psl - was mentioned as problematic, since it relies on some extensions which not everyone always has installed (e..g `bcmath`, `sodium`, etc.). This creates a chicken/egg problem if the extension someone wants to install is `bcmath`! Whilst I love the functionality we're using in `azjezz/psl` we may need to find another way; we're not using the `azjezz/psl` parts that use those extensions (we're ONLY using the nicely-typed JSON parsing, and only in a couple of places), and I don't think it's reasonable at all to ask the author that these parts are split off into a separate package.

ref: https://github.com/mongodb/mongo-php-driver/pull/1624#discussion_r1742527298

kynx:
@asgrim I’ve only just started following this repo and applaud your efforts. And yes, the fewer dependencies the better for a project that is going to bootstrap installs. I’d go further and say `symfony/console` is overkill for the simple CLI proposed - `getopts` ain’t much fun, but it works and is available on every install.

That said, the most important thing upfront is to get a functioning extension manager out the door and extension authors buying in. Don’t let dependency worries slow this down!

OT, but the big thing for me is to have tooling like renovate recognise when extensions need updating.

asgrim:
> to have tooling like renovate recognise when extensions need updating.

being able to identify which extensions are installed is definitely in the works (`pie show`), but I've added your suggestion to our design doc issue tracker - see ThePHPF/pie-design#24 - thank you @kynx !

asgrim:
[...]

PR #78, opened 2024-10-26
Replace internal NumberOfCores by CpuCoreCounter

theofidry:
[CpuCoreCounter](https://github.com/theofidry/cpu-core-counter) is the de-facto standard to get either the number of cores, or, more recently, the number of cores available for parallelization. You can see https://github.com/theofidry/cpu-core-counter/issues/11 for the motivations behind this library


asgrim:
I am torn about this; on one hand, I am a big fan of using libraries to avoid rehashing the same solution (and I love that this is a problem already solved!), and on another hand, we already experienced issues in the past about dependencies #31 which resulted in one dependency being removed. I'll have a think... :thinking:

theofidry:
@asgrim sorry didn't see your reply.

I think it's very reasonable to be wary of dependencies as a rule, and for a project that is a lot more critical such as pie/Composer, it is more true than ever. But like for Composer's XdebugHandler, there is still value in packages, even if sometimes they are very simple (code-wise).

So whilst there is ways to get away with a cheap code, properly fixing this problem (the CPU core count), is much harder. When I created that repo, I expect nothing more than a copy/paste of the PHPStan code. But as usual, you dig a bit deeper, and you find there is more cases & problems (similarly XdebugHandler did grow quite a bit – for the better, from its initial code).

I agree it is very important that this project has good control on the dependencies. So if needed I'm happy to both give control as a "backup" and adapt some constraints if necessary (FYI it already does some heavy compromises such as being PHP 7.2 compatible for PHPStan).

asgrim (approved):

asgrim:
Thanks @theofidry !

Issue #80, opened 2024-10-29
`pie build` fails on FreeBSD


TimWolla:
```
php bin/pie build -vvvv asgrim/example-pie-extension
```

results in:

```
In Process.php line 270:
                                                                                                        
  [Symfony\Component\Process\Exception\ProcessFailedException]                                          
  The command "'make' '--jobs' '2'" failed.                                                             
                                                                                                        
  Exit Code: 2(Misuse of shell builtins)                                                                
                                                                                                        
  Working directory: /tmp//pie_downloader_6720c5c851ae13.12531997/asgrim-example-pie-extension-f9ed13e  
                                                                                                        
  Output:                                                                                               
  ================                                                                                      
                                                                                                        
                                                                                                        
  Error Output:                                                                                         
  ================                                                                                      
  make: illegal argument to -j -- must be positive integer!                                                                                                             
```

Explicitly specifying `-j 1` works.

```
php bin/pie build -j 1 -vvvv asgrim/example-pie-extension
```

PR #81, opened 2024-10-29
Fix building with multiple jobs for BSDmake

TimWolla:
BSDmake does not accept the `--jobs` long-form flag and requires the use of `-j`.

The issue can be reproduced on a Ubuntu docker container by installing `bmake` and invoking that instead of `make`.

Fixes php/pie#80


asgrim:
Thank you @TimWolla ! 🙌

PR #82, opened 2024-11-07
Add extension-src directive

zeriyoshi:
Add `extension-src` directive to the `php-ext` section in `composer.json`. 

This enables support for patterns where extension source code is not located in the project root, which is commonly seen in third-party extensions.

https://github.com/protocolbuffers/protobuf/tree/main/php/ext/google/protobuf
```
{
    "name": "google/protobuf",
    "type": "php-ext",
    "description": "protocolbuffers php extension",
    "php-ext": {
        "extension-name": "protobuf",
        "extension-src": "php/ext/google/protobuf"
    }
}
```


zeriyoshi:
I don't think this test failure was caused by the change, but I'm not familiar enough with Windows to know...
https://github.com/php/pie/actions/runs/11720819724/job/32646969527?pr=82#step:6:25

asgrim:
@zeriyoshi thank you for this! However, I wish you had reported an issue first to discuss the requirement :sweat_smile: a lot of this code is currently being reworked at the moment to use a new pattern using the Composer Installer; it is absolutely feasible for us to accommodate this, but I'll need to take this on and rework this :grin:

zeriyoshi:
@asgrim 
My apologies! I thought if you were going to make a suggestion, it would be better to have an example first (bad habit of mine).
I'm glad to know what's going on! Thank you very much!

asgrim:
Re-opened PR #115 to cover this - thanks @zeriyoshi !

PR #83, opened 2024-11-07
Adapt to use composer installer

### LINKED ISSUE #48: Investigate improving internals with Composer Installer
I had previously discussed with Jordi about using the Composer Installer system to manage download/build/install process. I'd done some previous investigation work into this in a now outdated branch (https://github.com/asgrim/pie/tree/experiment-with-composer-installer) but at the time, it felt a bit of a rabbit hole. Now that we've got PIE approximately working, it would be worth revisiting this approach and take advantage of the Composer internals more.

asgrim:
Fixes #48


src/BinaryFile.php
```diff
@@ -0,0 +1,37 @@
...
+        public readonly string $checksum,
+    ) {
+    }
+
+    /** @param non-empty-string $filePath */
+    public static function fromFileWithSha256Checksum(string $filePath): self
+    {
+        return new self(
+            $filePath,
+            hash(self::HASH_TYPE_SHA256, file_get_contents($filePath)),
```
TimWolla, line 34: Could be `hash_file()` to not load the entire file contents into memory.
```suggestion
            hash_file(self::HASH_TYPE_SHA256, $filePath),
```

src/BinaryFile.php
```diff
@@ -0,0 +1,37 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Php\Pie;
+
+use function file_get_contents;
+use function hash;
```
TimWolla, line 8:
```suggestion
use function hash_file;
```

src/BinaryFile.php
asgrim: Thanks!

src/BinaryFile.php
```diff
@@ -0,0 +1,36 @@
...
+use function hash_file;
+
+/**
+ * @internal This is not public API for PIE, so should not be depended upon unless you accept the risk of BC breaks
+ *
+ * @immutable
+ */
+final class BinaryFile
+{
+    private const HASH_TYPE_SHA256 = 'sha256';
```
Seldaek, line 16: Why not sha3-256 for new stuff?

TimWolla: See https://news-web.php.net/php.internals/124694 and specifically https://news.ycombinator.com/item?id=14469614.

Also SHA-256 comes with hardware support for PHP 8.4, so it's really blazing fast (php/php-src#15152).

src/ComposerIntegration/ComposerIntegrationHandler.php
```diff
@@ -0,0 +1,78 @@
...
+            $composer->getRepositoryManager()->getLocalRepository()->removePackage($pkg);
+        }
+
+        $composerInstaller = Installer::create($this->arrayCollectionIo, $composer);
+        $composerInstaller
+            ->setAllowedTypes(['php-ext', 'php-ext-zend'])
+            ->setInstall(true)
+            ->setIgnoredTypes([])
+            ->setDryRun(false)
+            ->setDownloadOnly(false);
```
Seldaek, line 62: Nice! That's kinda what I envisioned when I added the allowed types stuff 👍🏻

src/BinaryFile.php
Seldaek: Ok, but if I read this correctly you have to do hash_hmac to prevent length extension attacks with SHA-256? How would that look here? Because in this file its just doing hash_file, is that safe then? I mean I don't think in this particular case the hashes are security-relevant, but just trying to understand.

TimWolla: > Ok, but if I read this correctly you have to do hash_hmac to prevent length extension attacks with SHA-256?

You read that correctly, but this attack vector is not relevant to this use case. To my understanding we care only about integrity here. Length extension attacks break authentication: The attacker knows a valid hash and the corresponding payload, but does not know the signature key.

To give a simplified example, if the signature is created as follows:

```
$hash = hash('sha256', $secretKey . $payload);
```

Then an attacker would be able to determine the following hash, even without knowing the `$secretKey`:

```
$hash = hash('sha256', $secretKey . $payload . $extension);
```

This would allow the attacker to pretend that `$payload . $extension` is validly signed. This can be relevant for file formats where the last entry wins: For example a tar archive.

src/ComposerIntegration/InstallAndBuildProcess.php
```diff
@@ -0,0 +1,82 @@
...
+    public function __invoke(
+        PartialComposer $composer,
+        PieComposerRequest $composerRequest,
+        CompletePackageInterface $composerPackage,
+        string $installPath,
+    ): void {
+        $output = $composerRequest->pieOutput;
+
+        $downloadedPackage = DownloadedPackage::fromPackageAndExtractedPath(
+            Package::fromComposerCompletePackage($composerPackage),
```
asgrim, line 36: Note to self: check into missing ext name issue:

```
 Php\Pie\ExtensionName::normaliseFromString() at /pwd/src/ExtensionName.php:68
 Php\Pie\ExtensionName::determineFromComposerPackage() at /pwd/src/DependencyResolver/Package.php:64
 Php\Pie\DependencyResolver\Package::fromComposerCompletePackage() at /pwd/src/ComposerIntegration/InstallAndBuildProcess.php:36
 Php\Pie\ComposerIntegration\InstallAndBuildProcess->__invoke() at /pwd/src/ComposerIntegration/PiePackageInstaller.php:60
```

src/BinaryFile.php
asgrim: > To my understanding we care only about integrity here

yep this is only for checking if the .so matches the one built by PIE (e.g. if it gets replaced externally), so doesn't really need to be secure as such, it is literally "has this thing changed" :)

src/ComposerIntegration/InstallAndBuildProcess.php
asgrim: Think the issue was out of date deps; https://github.com/composer/composer/commit/186d78cde397a3f4adbd600d61e5d4b85dbf7832 is needed for this branch to work; I do need to check if a tag has been released since :eyes:

asgrim: yes, we need min `^2.8.2` now

asgrim: Updated in b28d602

PR #88, opened 2024-11-14
Fallback on `PharData` to extract archives if `ext-zip` is unavailable

alexandre-daubois:
Came across this todo while looking for my first PR here. Let's use `PharData` if `ext-zip` is not available?


asgrim:
Thank you for this, but `ExtractZip` is actually removed completely in #83 :( sorry for wasting your time

alexandre-daubois:
No worries! I'll have a look at other todos and issues 😉

PR #90, opened 2024-11-14
Use Behat attributes instead of annotations

alexandre-daubois:
Attributes are available in Behat, I think they should be used instead of annotations.


asgrim (approved):

asgrim:
Thanks @alexandre-daubois

PR #91, opened 2024-11-14
Better authorization header handling in `AddAuthenticationHeader`

alexandre-daubois:
By limiting the number of results that `explode()` returns, we're fine dealing with values containing a `:`. Also, it ease the check of the parts validity as the result is way more deterministic compared to calling to `explode()` without limit.

### DIFF
--- a/src/Downloading/AddAuthenticationHeader.php
+++ b/src/Downloading/AddAuthenticationHeader.php
@@ -9,10 +9,11 @@ use Php\Pie\DependencyResolver\Package;
 use Psr\Http\Message\RequestInterface;
 use RuntimeException;
 
+use function array_map;
 use function array_walk;
+use function count;
 use function explode;
 use function sprintf;
-use function trim;
 
 /** @internal This is not public API for PIE, so should not be depended upon unless you accept the risk of BC breaks */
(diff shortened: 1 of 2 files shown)


asgrim (approved):

src/Downloading/AddAuthenticationHeader.php
```diff
@@ -27,9 +28,13 @@ public static function withAuthHeaderFromComposer(RequestInterface $request, Pac
         array_walk(
             $authHeaders,
             static function (string $v) use (&$request): void {
-                // @todo probably process this better
-                $headerParts = explode(':', $v);
-                $request     = $request->withHeader(trim($headerParts[0]), trim($headerParts[1]));
+                $headerParts = array_map('trim', explode(':', $v, 2));
```
faizanakram99, line 31: nit: `'trim'` can be replaced with `trim(...)`, first callable syntax is available since php 8.1

alexandre-daubois: That's true! But by using the "string syntax", the callable is resolved at the engine level instead of the language level which can be a bit faster. But if the project coding standards has a rule to use first class callable, let's update this part 🙂

PR #94, opened 2024-11-15
Remove `escapeshellarg()` quoting for configure flags

TimWolla:
This quoting is not necessary and in fact actively harmful, because the `./configure` command is not passed to a shell as a string where word-splitting happens, but as an argv array which the process will use as-is.

As an example, previously providing the `--with-library=/tmp/foo` flag would result in the `--with-library='/tmp/foo'` flag being passed as the argument, which then would be interpreted as if `"--with-library='/tmp/foo'"` was entered into a shell. In other words: The library was expected to be found in a directory literally starting with a single-quote.


asgrim:
Thanks @TimWolla !

PR #95, opened 2024-11-15
Use Composer HttpDownloader and remove Guzzle

### LINKED ISSUE #50: Switch from guzzle / psr7 to composer's HttpDownloader
Composer already has a battle-tested HTTP client built in.
What about leveraging it? That'd make the experience more consistent with composer itself.

asgrim:
Fixes #50


composer.json
```diff
@@ -30,8 +30,6 @@
         "php": "8.1.*||8.2.*||8.3.*||8.4.*",
         "ext-zip": "*",
         "composer/composer": "^2.8.2",
-        "guzzlehttp/guzzle": "^7.8",
-        "guzzlehttp/psr7": "^2.6",
         "illuminate/container": "^10.47",
         "psr/http-message": "^2.0",
```
stof, line 34: Shouldn't `psr/http-message` also be removed ? It is used by guzzle by I cannot find other usages

asgrim: Quite possibly; I will check the deps again now :)

PR #96, opened 2024-11-15
Add the `--with-phpize-path` option

alexandre-daubois:
Just like the `--with-php-config` and `--with-php-path` options, you can specify the path to `phpize` with this new option. If the option is not given, we fallback on the current `phpize` autodetection.

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -74,6 +74,13 @@ to run PIE, but we want to download the extension for PHP 8.3:
 > C:\php-8.1.7\php.exe C:\pie.phar install --with-php-path=C:\php-8.3.6\php.exe example/example-pie-extension
 ```
 
+You may also need to use the corresponding `phpize` command for the target PHP
+version, which can be specified with the `--with-phpize-path` option:
+
+```shell
+pie install --with-phpize-path=/usr/bin/phpize7.2 my/extension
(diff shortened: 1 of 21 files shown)


src/Platform/TargetPlatform.php
```diff
@@ -28,6 +29,7 @@ public function __construct(
         public readonly ThreadSafetyMode $threadSafety,
         public readonly int $makeParallelJobs,
         public readonly WindowsCompiler|null $windowsCompiler,
+        public readonly PhpizePath $phpizePath,
```
asgrim, line 32: Not sure about this change overall currently; we can't necessarily eagerly try to determine the `phpize` path; since Windows doesn't have phpize in the same way; so this will completely break on Windows.

alexandre-daubois: Indeed all Windows tests are broken right now. I'll try to figure out another solution

alexandre-daubois:
@asgrim I changed the approach. Now, the path is given to the PieComposerRequest and then to the Build as it is done for configure options. Is it better to you?

asgrim (approved):

asgrim:
Thanks @alexandre-daubois  yep this should help
!

PR #99, opened 2024-11-17
Update usage.md

oleg-andreyev:
updated download URL


asgrim:
Thanks; couple things 1) I'd like to avoid pointing the docs to a specific version, as if someone forgets to update the docs after a release, the docs will point to an old version 2) the [`latest` link](https://docs.github.com/en/repositories/releasing-projects-on-github/linking-to-releases#linking-to-the-latest-release) seems to only work for non pre-releases. I've changed it to not be a pre-release for now. This PR is no longer required

PR #102, opened 2024-11-18
Add build tools to Pie requirements in the docs

alexandre-daubois:
A few lines about packages required to build extensions as well as the commands to install those packages.


asgrim:
Thanks @alexandre-daubois !

Issue #87, opened 2024-11-12
Stick to architecture naming conventions on Windows


cmb69:
https://github.com/ThePHPF/pie-design?tab=readme-ov-file#windows-binaries mentions that the architecture should be `x86_64` for 64bit Windows versions, while so far we used `x64` everywhere (PHP on Windows binaries, PECL DLL packages etc.) I don't see any good reason to break with this convention.

cmb69:
> `x64` is really named incorrectly (it is a shorthand for `x86_64`).

Tell that Microsoft: https://learn.microsoft.com/en-us/cpp/build/how-to-enable-a-64-bit-visual-cpp-toolset-on-the-command-line?view=msvc-170. Or any Windows developer.

asgrim:
The way I read this [Wikipedia](https://en.wikipedia.org/wiki/X86-64#Industry_naming_conventions) suggests that `x86_64` (or, `x86-64`) is correct, and `x64` is a short hand that Windows uses:

> Since AMD64 and Intel 64 are substantially similar, many software and hardware products use one vendor-neutral term to indicate their compatibility with both implementations. AMD's original designation for this processor architecture, "x86-64", is still used for this purpose,[2] as is the variant "x86_64".[3][4] Other companies, such as Microsoft[6] and Sun Microsystems/Oracle Corporation,[5] use the contraction "x64" in marketing material. 

We are trying to ensure PIE is platform agnostic, which means the architecture naming should be common throughout. Making an exception for Windows because Microsoft decided to use a different name may create extra unnecessary maintenance burden.

This feedback is unfortunately very late in the game, and we did actually already co-ordinate on this issue and have this discussion some time ago, and agreed that we should be consistent with the architecture naming (thus, `x86_64`). It is for this reason that https://github.com/php/php-windows-builder/ GH action produces the binaries named with `x86_64`.

PR #107, opened 2024-11-19
Replace x86_64 by x64 in Windows documentation

macintoshplus:
For Windows, the arch code is `x86` for Intel 32 bit, `x64` for Intel 64 bit, and `arm64` for ARM 64 bit.

This is the extract from some PHP versions on Windows:

image

image

image


asgrim:
See #87 where this was discussed already

PR #111, opened 2024-11-19
Prevent the `--with-phpize-path` option to be used on Windows

alexandre-daubois:
Followup of #96

### DIFF
--- a/src/Command/CommandHelper.php
+++ b/src/Command/CommandHelper.php
@@ -113,6 +113,15 @@ final class CommandHelper
             throw new InvalidArgumentException('The --with-php-path=/path/to/php cannot be used on non-Windows, use --with-php-config=/path/to/php-config instead.');
         }
 
+        if (Platform::isWindows() && $input->hasOption(self::OPTION_WITH_PHPIZE_PATH)) {
+            /** @var mixed $withPhpizePath */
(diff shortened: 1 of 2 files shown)


src/Command/CommandHelper.php
```diff
@@ -113,6 +113,15 @@ public static function determineTargetPlatformFromInputs(InputInterface $input,
...
 
+        if (Platform::isWindows() && $input->hasOption(self::OPTION_WITH_PHPIZE_PATH)) {
+            /** @var mixed $withPhpizePath */
+            $withPhpizePath = $input->getOption(self::OPTION_WITH_PHPIZE_PATH);
+
+            if (is_string($withPhpizePath) && trim($withPhpizePath) !== '') {
+                throw new InvalidArgumentException('The --with-phpize-path=/path/to/phpize cannot be used on Windows.');
+            }
+        }
+
```
kynx, line 124: Random thought here - I'm just following this repo out of enthusiastic interest 😄 - but would it make more sense to check for windows before adding the option in the first place? It'd be less frustrating for users than error messages...

alexandre-daubois: Thought the same! I'm not sure which is best: having options available on some platforms and not on others, or having the same options everywhere even if it cannot be used directly?

For the sake of consistency and UX, I would say the second solution which is more verbose and explicit to the end user 🤔

asgrim (approved):

PR #112, opened 2024-11-19
Use `#[RequiresOperatingSystemFamily]` where possible

alexandre-daubois:
It is easier to find all tests where Windows is required, as `self::markTestSkipped()` messages are not 100% consistent across all calls. Also it should be better for PHPUnit so it doesn't have to start executing the test to skip it.


asgrim (approved):

asgrim:
Thanks @alexandre-daubois

PR #113, opened 2024-11-19
Add Homebrew TAP support

stloyd:
This requires adding a new secret to actions `ACCESS_TOKEN` with "repo" & "workflow" scopes.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -50,3 +50,21 @@ jobs:
           files: |
             pie.phar
             pie.phar.asc
+      - name: Get release tag
+        run: |
+          RELEASE_TAG=${GITHUB_REF/refs\/tags\//}
+          echo "RELEASE_TAG=$RELEASE_TAG" >> $GITHUB_ENV
+      - name: Update Homebrew formula
+        uses: mislav/bump-homebrew-formula-action@v3
+        with:
+          formula-name: php-pie
+          formula-path: homebrew/Formula/php-pie.rb
+          homebrew-tap: php/pie
+          download-url: https://github.com/php/pie/releases/download/${{ env.RELEASE_TAG }}/pie.phar
+          commit-message: |
+            Update Homebrew TAP formula: {{formulaName}} to version: {{version}}
+
+            Commit executed by automatic flow after every release
+        env:
+          # the personal access token should have "repo" & "workflow" scopes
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -50,3 +50,21 @@ jobs:
           files: |
             pie.phar
             pie.phar.asc
+      - name: Get release tag
+        run: |
+          RELEASE_TAG=${GITHUB_REF/refs\/tags\//}
+          echo "RELEASE_TAG=$RELEASE_TAG" >> $GITHUB_ENV
+      - name: Update Homebrew formula
+        uses: mislav/bump-homebrew-formula-action@v3
```
asgrim, line 58: I don't really understand what this is for / what this is doing? I'm assuming something to do with brew, but it's something I'm unfamiliar with. Can you give me a TLDR? or point me to some docs to understand this process please?

homebrew/Formula/php-pie.rb
```diff
@@ -0,0 +1,17 @@
+class PhpPie < Formula
+  desc "PIE - the PHP Installer for Extensions"
+  homepage "https://github.com/php/pie"
+  url "https://github.com/php/pie/releases/download/0.2.0/pie.phar"
```
asgrim, line 4: We should not hard-code this, the release process should be automated, and this will **definitely** be forgotten if it is not automated :)

.github/workflows/release.yml
```diff
@@ -50,3 +50,21 @@ jobs:
...
+        with:
+          formula-name: php-pie
+          formula-path: homebrew/Formula/php-pie.rb
+          homebrew-tap: php/pie
+          download-url: https://github.com/php/pie/releases/download/${{ env.RELEASE_TAG }}/pie.phar
+          commit-message: |
```
asgrim, line 64: Is this making a commit to `php/pie`? :thinking:

homebrew/Formula/php-pie.rb
theofidry: Personally I went with making a separate repo for this, which would be expected to be written on by a bothttps://github.com/box-project/homebrew-box

pronskiy: Could be https://github.com/php/pie/releases/latest/download/pie.phar

PR #114, opened 2024-11-19
Make all PHPUnit data providers static

alexandre-daubois:
Non-static data providers is deprecated since PHPUnit 10.


asgrim (approved):

asgrim:
Thanks @alexandre-daubois

PR #116, opened 2024-11-20
Add support for `os-families-exclude` in extensions `composer.json`

### LINKED ISSUE #106: How to require an specific OS
The Win32Service PHP extension can be built and used only on the Microsoft Windows operating system.

How to specify this constraint on `composer.json` file?

alexandre-daubois:
Fix #106

Allows to specify OS families which the extension is incompatible with:

```json5
{
    "name": "myvendor/myext",
    "php-ext": {
        "extension-name": "myext",
        "configure-options": [
            // ...
        ],
        "os-families-exclude": ['windows', 'darwin']
    }
}
```

When the syntax is validated, I'll create a PR in Composer to allow the new key in the `Package::$phpExt` property.

Related to https://github.com/ThePHPF/pie-design/pull/29


src/Command/CommandHelper.php
```diff
@@ -210,6 +212,21 @@ public static function bindConfigureOptionsFromPackage(Command $command, Package
         self::validateInput($input, $command);
     }
 
+    public static function ensureOsFamilyCompatibilityFromPackage(Package $package): void
```
asgrim, line 215: This should be done in `\Php\Pie\DependencyResolver\ResolveDependencyWithComposer` - example in `\Php\Pie\DependencyResolver\ResolveDependencyWithComposer::assertCompatibleThreadSafetyMode`

src/Command/CommandHelper.php
```diff
@@ -210,6 +212,21 @@ public static function bindConfigureOptionsFromPackage(Command $command, Package
...
+    {
+        if ($package->excludedOsFamilies === []) {
+            return;
+        }
+
+        if (in_array(PHP_OS_FAMILY, $package->excludedOsFamilies, true)) {
```
asgrim, line 221: This **must** use the `TargetPlatform`; since users of PIE can install a package for a different version of PHP than the one running PIE

src/DependencyResolver/Package.php
```diff
@@ -25,7 +26,10 @@
  */
 final class Package
 {
-    /** @param list<ConfigureOption> $configureOptions */
+    /**
+     * @param list<ConfigureOption> $configureOptions
+     * @param list<string>          $excludedOsFamilies
```
asgrim, line 31: should be a `list<OsFamily>` and create a new `enum` for this with the allowed values:

 * Windows
 * BSD
 * Darwin
 * Solaris
 * Linux
 * Unknown

ref: https://www.php.net/manual/en/reserved.constants.php#constant.php-os-family

asgrim (changes requested):

src/Command/CommandHelper.php
alexandre-daubois: That's what I thought first. Does it mean that user can even build and install extensions for a different OS than the one running Pie with a cross compilation system?

asgrim: In practice, at the moment, no. Theoretically, it would be possible to change various command invocations and file operations to run against a remote PHP install. But, that hasn't been discussed though as whether it's something we even want to do.

alexandre-daubois: Right thanks, I'll do the update 👍

alexandre-daubois:
Revamped the whole PR to address your comments.
- Things are done in `ResolveDependencyWithComposer` instead of `CommandHelper`
- An enum holding OS families is added
- It is added to the TargetPlatform
- `PhpBinaryPath` is used to determine the OS family, just like the OS

tillkruss:
This is great, love to exclude windows.

src/DependencyResolver/Package.php
```diff
@@ -25,7 +30,11 @@
  */
 final class Package
 {
-    /** @param list<ConfigureOption> $configureOptions */
+    /**
+     * @param list<ConfigureOption>       $configureOptions
+     * @param list<OperatingSystemFamily> $compatibleOsFamilies
```
asgrim, line 35: Hey @alexandre-daubois please can you check out some suggested improvements alexandre-daubois/pie#2

alexandre-daubois:
Remaining CS warnings should be fixed once https://github.com/composer/composer/pull/12218 is merged

asgrim:
PR composer/composer#12218 is now merged; please can you update the composer req, that should fix the static analysis

alexandre-daubois:
Here you go, all green! 🙂

asgrim (approved):
Thanks @alexandre-daubois

Seldaek:
It's now tagged as part of composer 2.8.4

alexandre-daubois:
Thanks!

PR #118, opened 2024-11-20
Add the `--dry-run` option

### LINKED ISSUE #18: Add `--dry-run` option
The design doc specifies a `--dry-run` option, which allows simulation of a download/build without actually doing it.

https://github.com/ThePHPF/pie-design?tab=readme-ov-file#--dry-run

alexandre-daubois:
Fix #18

Allows using the `--dry-run` option with both `build` and `install` command. This wasn't added the `download` command, as I'm not really sure it makes sense to have it.

### DIFF
--- a/src/Command/CommandHelper.php
+++ b/src/Command/CommandHelper.php
@@ -38,6 +38,7 @@ final class CommandHelper
     private const OPTION_WITH_PHP_PATH              = 'with-php-path';
     private const OPTION_WITH_PHPIZE_PATH           = 'with-phpize-path';
     private const OPTION_MAKE_PARALLEL_JOBS         = 'make-parallel-jobs';
+    private const OPTION_DRY_RUN                    = 'dry-run';
 
     /** @psalm-suppress UnusedConstructor */
     private function __construct()
@@ -87,6 +88,16 @@ final class CommandHelper
          * Note, this means you probably need to call {@see self::validateInput()} to validate the input manually...
          */
         $command->ignoreValidationErrors();
+    }
+
+    public static function configureDryRunOption(Command $command): void
+    {
+        $command->addOption(
+            self::OPTION_DRY_RUN,
+            null,
+            InputOption::VALUE_NONE,
+            'Do not actually build or install the extension, just show the steps that would be run.',
+        );
     }
 
     public static function validateInput(InputInterface $input, Command $command): void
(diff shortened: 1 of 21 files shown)


src/Command/CommandHelper.php
```diff
@@ -89,6 +90,16 @@ public static function configureDownloadBuildInstallOptions(Command $command): v
         $command->ignoreValidationErrors();
     }
 
+    public static function configureDryRunOption(Command $command): void
```
asgrim, line 93: IMO this should be merged into `configureDownloadBuildInstallOptions`

asgrim (changes requested):

src/Command/CommandHelper.php
alexandre-daubois: I'm not sure, because it would mean the `download` command could dry run. I'm not sure this really makes sense to dry run a download?

alexandre-daubois:
Thanks James, I addressed all your comments! I really like the `dry-run::` and the fact it could be used later to check whether a dry run was launched or not.

src/Command/CommandHelper.php
asgrim: I think that is the part that is confusing; whilst we do actually need to download to dry-run the build+install parts, we should be able to restore the `PIE_WORKING_DIRECTORY` to its previous state.

Running `\Php\Pie\ComposerIntegration\ComposerIntegrationHandler::__invoke` updates your `${PIE_WORKING_DIRECTORY}/pie.json` (+`${PIE_WORKING_DIRECTORY}/pie.lock`) and subsequently updates `{$PIE_WORKING_DIRECTORY}/vendor/...` with the new install.

The correct approach for dry-run would be:

 - Resolve dependency
 - Back up `pie.json`+`pie.lock`+`vendor`
   - Run the Composer integration handler
   - If requested, run the build with dry run flag
   - If requested, run the install with dry run flag
 - Restore the backups

BUT - perhaps a simpler way of doing this however, would be to sandbox the dry-run download by adding `dry_run` prefix or similar to `$targetPlatformPath` in `\Php\Pie\Platform::getPieWorkingDirectory`, so it is downloaded in isolation from someone's "real" `PIE_WORKING_DIRECTORY`; then after the dry-run download/build/install is complete, remove the pie working directory (checking it is only for a dry-run).

FWIW, this is one of the reasons I'd been avoiding implementing `--dry-run`; it is messy and non-trivial to implement correctly :laughing:

alexandre-daubois: Thank you for this complete insight! I think I'm going to close this PR. If we were to implement it, I think it should be started again from scratch, because the current implementation here is not what we want.

I'll link your comment in the issue for when I'll have a bit more time for this feature, or if someone wants to take over.

Thanks again!

alexandre-daubois:
Closed as explained in https://github.com/php/pie/pull/118#discussion_r1888276217

PR #119, opened 2024-11-20
Add Dependabot for GitHub Actions

TimWolla:
The release.yml workflow currently emits a warning:

> The following actions use a deprecated Node.js version and will be forced to
> run on node20: softprops/action-gh-release@v1.

A newer version for that action is available. Dependabot should make it much easier to keep the actions up to date.


asgrim (approved):

asgrim:
Thanks @TimWolla !

PR #121, opened 2024-11-21
Run `composer bump`

TimWolla:
Given that PIE is a project rather than a library and is distributed as a Phar archive, we do not need to keep broad compatibility with older library versions. Run `composer bump` to prevent accidental downgrades of libraries, by increasing the versions in `composer.json` to match the currently locked versions in `composer.lock`.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -52,6 +52,7 @@
     "config": {
         "sort-packages": true,
         "preferred-install": "dist",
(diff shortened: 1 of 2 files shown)


asgrim (approved):

asgrim:
Thanks @TimWolla !

PR #122, opened 2024-11-21
feat: CI job for building Docker binary-only images

### LINKED ISSUE #137: PIE installable in a Dockerfile?
Relates to existing PR php/pie#122

Should PIE be installable in a Dockerfile?

```shell
COPY --from=ghcr.io/php/pie:latest /pie /usr/bin/pie
```

Points to note:

 - This is NOT an executable image (i.e. you would not be able to run `docker run --rm ghcr.io/php/pie:latest`) - this is only publishing the PHAR but contained within the Dockerfile.
 - Alternate approaches would be using `ADD `, for example (note; untested) `ADD https://github.com/php/pie/releases/latest/download/pie.phar /usr/bin/pie`
 - There is an open question about attestation; how to guarantee the image. One approach could be to use `gh attestation verify` (recent example for the PHAR itself in #128 )

Currently the only supported distribution mechanism is downloading the PHAR from the https://github.com/php/pie/releases page. Other distribution mechanisms should be carefully evaluated on a case-by-case basis.

Wirone:
Similar to composer/docker#250 ([more info](https://blog.codito.dev/2022/11/composer-binary-only-docker-images/)), this PR introduces a CI process for releasing binary-only images for Pie 🥳!

You can see the result [here](https://github.com/wirone/php-pie/pkgs/container/php-pie), there is an image in my fork's Package Registry, built [here](https://github.com/Wirone/php-pie/actions/runs/12083428508/job/33696550468).

Here's how the image looks like (`dive ghcr.io/wirone/php-pie:latest-bin`):
image

~~PS: I decided to not verify GPG signature during build because we can trust our own workflow (we copy PHAR file built in previous stage), but also I had some problems with verification (GPG imported correctly, but verification failed). Users may verify the file on their own, if they want, using attached `pie.asc`.~~ Signing was dropped.

Fixes #137

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -57,7 +57,7 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v4
 
-      - name: Restore built PHAR
+      - name: Fetch built PHAR from artifacts
         uses: actions/download-artifact@v4
         with:
           name: pie-${{ github.sha }}.phar
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -6,10 +6,7 @@ on:
       - published
 
 permissions:
-  attestations: write
   contents: read
-  id-token: write
-  packages: write
 
 jobs:
   build-phar:
@@ -47,6 +44,14 @@ jobs:
     name: Docker binary-only image
     runs-on: ubuntu-latest
     if: ${{ startsWith(github.ref, 'refs/tags/') }}
+
+    permissions:
+      # attestations:write is required to publish provenance attestations for the built Docker image
+      attestations: write
+      # id-token:write is required for logging in to the Docker registry
+      id-token: write
+      # packages:write is required to publish Docker images to GitHub's registry
+      packages: write
 
     steps:
       - name: Checkout repository
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -86,13 +86,12 @@ jobs:
         with:
           flavor: |
             latest=false
-            suffix=-bin
           images: ghcr.io/${{ github.repository }}
           tags: |
-            type=raw,value=latest
-            type=semver,pattern={{version}}
-            type=semver,pattern={{major}}.{{minor}}
-            type=semver,pattern={{major}}
+            type=raw,value=bin
+            type=semver,pattern={{version}}-bin
+            type=semver,pattern={{major}}.{{minor}}-bin
+            type=semver,pattern={{major}}-bin
 
       - name: Build and push Docker image
         id: build-and-push
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -7,6 +7,7 @@ on:
 
 permissions:
   contents: read
+  packages: write
 
 jobs:
   build-phar:
@@ -38,3 +39,56 @@ jobs:
         if: ${{startsWith(github.ref, 'refs/tags/') }}
         with:
           files: pie.phar
+
+  docker-binary-only-image:
+    needs: build-phar
+    name: Docker binary-only image
+    runs-on: ubuntu-latest
+    if: ${{ startsWith(github.ref, 'refs/tags/') }}
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Restore built PHAR
+        uses: actions/download-artifact@v4
+        with:
+          name: pie-${{ github.sha }}.phar
+
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v3
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+
+      - name: Log in to the Container registry
+        uses: docker/login-action@v3
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Extract metadata (tags, labels) for Docker
+        id: meta
+        uses: docker/metadata-action@v5
+        with:
+          flavor: |
+            latest=false
+            suffix=-bin
+          images: ghcr.io/${{ github.repository }}
+          tags: |
+            type=raw,value=latest
+            type=semver,pattern={{version}}
+            type=semver,pattern={{major}}.{{minor}}
+            type=semver,pattern={{major}}
+
+      - name: Build and push Docker image
+        uses: docker/build-push-action@v5
+        with:
+          context: .
+          platforms: linux/amd64,linux/arm64
+          file: Dockerfile
+          target: standalone-binary
+          push: true
+          tags: ${{ steps.meta.outputs.tags }}
+          labels: ${{ steps.meta.outputs.labels }}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -55,6 +55,11 @@ jobs:
         with:
           name: pie-${{ github.sha }}.phar
 
+      - name: Verify the PHAR
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: gh attestation verify pie.phar --repo ${{ github.repository }}
+
       - name: Set up QEMU
         uses: docker/setup-qemu-action@v3
 
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -46,11 +46,11 @@ jobs:
     if: ${{ startsWith(github.ref, 'refs/tags/') }}
 
     permissions:
-      # attestations:write is required to publish provenance attestations for the built Docker image
+      # attestations:write is required for build provenance attestation.
       attestations: write
-      # id-token:write is required for logging in to the Docker registry
+      # id-token:write is required for build provenance attestation.
       id-token: write
-      # packages:write is required to publish Docker images to GitHub's registry
+      # packages:write is required to publish Docker images to GitHub's registry.
       packages: write
 
     steps:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -87,11 +87,12 @@ jobs:
           flavor: |
             latest=false
           images: ghcr.io/${{ github.repository }}
+          # @TODO v1.0 Consider introducing more granular tags (major and major.minor)
+          # @see https://github.com/php/pie/pull/122#pullrequestreview-2477496308
+          # @see https://github.com/php/pie/pull/122#discussion_r1867331273
           tags: |
             type=raw,value=bin
             type=semver,pattern={{version}}-bin
-            type=semver,pattern={{major}}.{{minor}}-bin
-            type=semver,pattern={{major}}-bin
 
       - name: Build and push Docker image
         id: build-and-push
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -6,7 +6,9 @@ on:
       - published
 
 permissions:
+  attestations: write
   contents: read
+  id-token: write
   packages: write
 
 jobs:
@@ -88,6 +90,7 @@ jobs:
             type=semver,pattern={{major}}
 
       - name: Build and push Docker image
+        id: build-and-push
         uses: docker/build-push-action@v5
         with:
           context: .
@@ -97,3 +100,10 @@ jobs:
           push: true
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
+
+      - name: Generate artifact attestation
+        uses: actions/attest-build-provenance@v1
+        with:
+          subject-name: ghcr.io/${{ github.repository }}
+          subject-digest: ${{ steps.build-and-push.outputs.digest }}
+          push-to-registry: true
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,4 @@
+FROM scratch AS standalone-binary
+
+# @TODO change to --chmod=+x when https://github.com/moby/buildkit/pull/5380 is released
+COPY --chmod=0755 pie.phar /pie
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -35,7 +35,7 @@ PIE is published as binary-only Docker image, so you can install it easily durin
 COPY --from=ghcr.io/php/pie:bin /pie /usr/bin/pie
 ```
 
-Instead of `bin` tag (which represents latest binary-only image) you can also use explicit versions like `x.y.z-bin`, `x.y-bin` or `x-bin`, depending on stability level you want to achieve.
+Instead of `bin` tag (which represents latest binary-only image) you can also use explicit version (in `x.y.z-bin` format). Use [GitHub registry](https://ghcr.io/php/pie) to find available tags.
 
 > [!IMPORTANT]  
 > Binary-only images don't include PHP runtime so you can't use them for _running_ PIE. This is just an alternative way of distributing PHAR file, you still need to satisfy PIE's runtime requirements on your own.
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -27,6 +27,19 @@ system:
 sudo curl -L --output /usr/local/bin/pie https://github.com/php/pie/releases/latest/download/pie.phar && sudo chmod +x /usr/local/bin/pie
 ```
 
+### Docker installation
+
+PIE is published as binary-only Docker image, so you can install it easily during your Docker build:
+
+```Dockerfile
+COPY --from=ghcr.io/php/pie:latest-bin /pie /usr/bin/pie
+```
+
+Instead of `latest` you can also use explicit versions like `x.y.z-bin`, `x.y-bin` or `x-bin`, depending on stability level you want to achieve.
+
+> [!IMPORTANT]  
+> Binary-only images don't include PHP runtime so you can't use them for _running_ PIE. This is just an alternative way of distributing PHAR file, you still need to satisfy PIE's runtime requirements on your own.
+
 ## Prerequisites for PIE
 
 Running PIE requires PHP 8.1 or newer. However, you may still use PIE to install
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -32,10 +32,10 @@ sudo curl -L --output /usr/local/bin/pie https://github.com/php/pie/releases/lat
 PIE is published as binary-only Docker image, so you can install it easily during your Docker build:
 
 ```Dockerfile
-COPY --from=ghcr.io/php/pie:latest-bin /pie /usr/bin/pie
+COPY --from=ghcr.io/php/pie:bin /pie /usr/bin/pie
 ```
 
-Instead of `latest` you can also use explicit versions like `x.y.z-bin`, `x.y-bin` or `x-bin`, depending on stability level you want to achieve.
+Instead of `bin` tag (which represents latest binary-only image) you can also use explicit versions like `x.y.z-bin`, `x.y-bin` or `x-bin`, depending on stability level you want to achieve.
 
 > [!IMPORTANT]  
 > Binary-only images don't include PHP runtime so you can't use them for _running_ PIE. This is just an alternative way of distributing PHAR file, you still need to satisfy PIE's runtime requirements on your own.


.github/workflows/release.yml
```diff
@@ -50,3 +54,59 @@ jobs:
...
+          tags: |
+            type=semver,pattern={{version}}
+            type=semver,pattern={{major}}.{{minor}}
+            type=semver,pattern={{major}}
+
+      - name: Build and push Docker image
+        uses: docker/build-push-action@v5
+        with:
+          context: .
+          platforms: linux/amd64,linux/arm64
```
asgrim, line 105: Is there a way we can build this architecture-agnostic (rather than explicitly building for arm64/amd64/etc) :thinking:

docs/usage.md
```diff
@@ -29,6 +29,20 @@ system:
...
+
+```Dockerfile
+FROM php:8
+
+COPY --form=ghcr.io/php/pie:latest /pie /usr/bin/pie
+```
+
+Instead of `latest` you can also use explicit versions like `x.y.z`, `x.y` or `x`, depending on stability level you want to achieve.
+
+Each image contains `/pie.asc` file too, so you can verify GPG signature of bundled PHAR file (`/pie`).
```
TimWolla, line 44: The /pie.asc is exactly as (un)trustworthy as the binary itself, because it is downloaded from the same location.

TimWolla (changes requested):

Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM curlimages/curl AS phar-downloader
```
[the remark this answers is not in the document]
TimWolla: That should read `bookworm`, of course. The `bookwork-slim` variant ideally.

.github/workflows/release.yml
```diff
@@ -50,3 +54,59 @@ jobs:
...
+    permissions:
+      contents: read
+      packages: write
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v3
```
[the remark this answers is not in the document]
Wirone: When one does `COPY --from=ghcr.io/php/pie`, expects image for architecture currently image is built, if we build one arch then build on other archs will trigger warning AFAIK.

docs/usage.md
```diff
@@ -29,6 +29,20 @@ system:
 sudo curl -L --output /usr/local/bin/pie https://github.com/php/pie/releases/latest/download/pie.phar && sudo chmod +x /usr/local/bin/pie
 ```
 
+### Docker installation
+
+Pie is published as binary-only Docker image, so you can install it easily during build:
+
+```Dockerfile
+FROM php:8
```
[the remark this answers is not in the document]
Wirone: No, because it uses `8.1.*`, while `8` currently resolves to `8.4`. Maybe we should remove this and only document `COPY --from`?

.github/workflows/release.yml
Wirone: As pointed out [here](https://github.com/php/pie/pull/122#discussion_r1853101732), we can't do it (or at least I don't know how). Is it a problem, though?

Dockerfile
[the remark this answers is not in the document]
Wirone: You don't trust cURL team? These images are officially built from [`curl/curl-container`](https://github.com/curl/curl-container). What's the point of using Debian (`debian:bookworm-slim` is 97MB), when this image is 7x lighter (13MB)?

Dockerfile
```diff
@@ -0,0 +1,12 @@
...
+
+ARG PIE_VERSION
+
+RUN curl -L --output /tmp/pie https://github.com/php/pie/releases/download/${PIE_VERSION}/pie.phar \
+    && chmod +x /tmp/pie \
+    && curl -L --output /tmp/pie.asc https://github.com/php/pie/releases/download/${PIE_VERSION}/pie.phar.asc
```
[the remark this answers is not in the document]
Wirone: Let's resolve this [here](https://github.com/php/pie/pull/122#discussion_r1852966431).

.github/workflows/release.yml
```diff
@@ -50,3 +54,59 @@ jobs:
...
+        id: meta
+        uses: docker/metadata-action@v5
+        with:
+          flavor: |
+            latest=true
+          images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME_LOWER }}
+          tags: |
+            type=semver,pattern={{version}}
+            type=semver,pattern={{major}}.{{minor}}
+            type=semver,pattern={{major}}
```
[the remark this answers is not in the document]
Wirone: This is standard practice for `php` images too, you can pull `8.4.1`, `8.4` or just `8`. Composer and PHP-CS-Fixer also have this. I don't agree with your comment, this works as expected and offers 3 levels of stability, so users can choose how they want install the tool.

.github/workflows/release.yml
```diff
@@ -50,3 +54,59 @@ jobs:
...
+      - name: Log in to the Container registry
+        uses: docker/login-action@v3
+        with:
+          registry: ${{ env.REGISTRY }}
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Make image lowercase
+        run: |
+          echo IMAGE_NAME_LOWER=$(echo ${{ env.IMAGE_NAME }} | tr '[:upper:]' '[:lower:]') >> ${GITHUB_ENV}
```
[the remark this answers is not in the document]
Wirone: What do you mean? I think this is not needed here as the repo itself is named `php/pie`, but in general image names must be in lowercase, so this is a guard rail to keep it compatible regardless of repo name.

docs/usage.md
Wirone: What do you propose? Removing this completely and if anyone wants to verify it, should download ASC file from Github? This would be tricky with `X` and `X.Y` version binding, because you will download dynamic version.

.github/workflows/release.yml
asgrim: I was hoping you could build for `noarch` architecture or something to indicate that a single manifest can operate on any platform, since the only content of the image (the PHAR) is entirely platform agnostic. Not a major problem, but it creates overhead for something that should otherwise be a single step >.<

docs/usage.md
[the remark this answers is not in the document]
asgrim: I think that would be reasonable

.github/workflows/release.yml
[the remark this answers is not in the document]
asgrim: The repo name is indeed already lowercase, so this step can be removed.

.github/workflows/release.yml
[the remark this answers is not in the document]
asgrim: I think this is fine as is, if people want to use 1.2.3, or latest, or 1, or 1.2, they can. It is a reasonable trade-off IMO. With a tool like this, and Composer, *typically* (if you're working on modern applications), the latest is fine, but this allows people to pin to specific dependencies. Issues come with re-tagging (which I would strongly protest against, causes more issues than good), or retroactively tagging earlier releases, but this is very edge case.

docs/usage.md
asgrim: Hm; yes, I see Tim's point; it probably makes sense to remove the signature from the image and to point people to the GH releases list (which is where the signature is canonically kept).

.github/workflows/release.yml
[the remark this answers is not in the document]
asgrim: Closing, relevant discussion: https://github.com/php/pie/pull/122/files#r1853416941

.github/workflows/release.yml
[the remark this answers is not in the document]
Wirone: > Issues come with re-tagging (which I would strongly protest against, causes more issues than good), or retroactively tagging earlier releases, but this is very edge case.

This 💯. With proper release cycle it just works. Even such edge case can be covered by running the workflow from the Github UI with explicit params, but I believe it won't be needed.

Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM curlimages/curl AS phar-downloader
+
+ARG PIE_VERSION
+
+RUN curl -L --output /tmp/pie https://github.com/php/pie/releases/download/${PIE_VERSION}/pie.phar \
```
asgrim, line 5: Rather than downloading from GH, could we not use an artifact from the build step perhaps? :thinking:

Wirone: I've added steps for storing and retrieving artifacts, I'll test it soon on my fork.

docs/usage.md
Wirone: I removed `asc` file from the image and changed usage docs.

Dockerfile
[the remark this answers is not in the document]
Wirone: Outdated, since cURL is not needed anymore when PHAR is bundled into Docker image using artifact from previous build stage.

Dockerfile
Wirone: It works 👍😁. You can see it [here](https://github.com/Wirone/php-pie/actions/runs/11969197461) - artifact is created in `release-phar` job, then restored in `Docker binary-only image`. Image contains only PHAR now, I made it executable (with a `@TODO` waiting for [this](https://github.com/moby/buildkit/pull/5380) to be released).

image

Dockerfile
Wirone, line 1: FYI: this is currently not much usable for local development, but _if_ you plan to define Docker setup for working with the project, you just need to utilise multi-stage build (keep `standalone-binary` target for releases and add new one(s) for development).

docs/usage.md
```diff
@@ -29,6 +29,18 @@ system:
...
+
+PIE is published as binary-only Docker image, so you can install it easily during your Docker build:
+
+```Dockerfile
+COPY --form=ghcr.io/php/pie:latest /pie /usr/bin/pie
+```
+
+Instead of `latest` you can also use explicit versions like `x.y.z`, `x.y` or `x`, depending on stability level you want to achieve.
+
+If you want to verify signature of the PHAR, you need to additionally download `*.asc` file from [corresponding release](https://github.com/php/pie/releases).
```
TimWolla, line 42: As part of my PRs #127 and #128 to do some preliminary GitHub Action cleanup, I've also tried to run pie.phar in a clean `php:8.3-cli` Docker container. This failed, because `ext-zip` is not installed by default.

Given that information, I'm wondering it shipping PIE as proposed in this PR is a value-add for the user. Perhaps the Docker image should instead contain a “PIE installer”, which:

1. `docker-php-ext-install zip` according to best practices (incl. the necessary cleanup of build dependencies) with some smartness to handle both Debian and Alpine images.
2. Places PIE into `/usr/bin/pie` after appropriately verifying the signature.

Usage would then look something like this:

```
COPY --from=ghcr.io/php/pie:latest /usr/bin/install-pie
RUN /usr/bin/install-pie
```

Alternatively the Docker image contain `pie.phar` embedded into a wrapper script that checks the prerequisites, installs them when necessary and then executes pie.phar.

docs/usage.md
```diff
@@ -29,6 +29,18 @@ system:
 sudo curl -L --output /usr/local/bin/pie https://github.com/php/pie/releases/latest/download/pie.phar && sudo chmod +x /usr/local/bin/pie
 ```
 
+### Docker installation
+
+PIE is published as binary-only Docker image, so you can install it easily during your Docker build:
+
+```Dockerfile
+COPY --form=ghcr.io/php/pie:latest /pie /usr/bin/pie
```
TimWolla, line 37:
```suggestion
COPY --from=ghcr.io/php/pie:latest /pie /usr/bin/pie
```

.github/workflows/release.yml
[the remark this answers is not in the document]
TimWolla: > the latest is fine, but this allows people to pin to specific dependencies

Yes. But I'm arguing to think hard about whether we want to encourage users to pin to PIE 1.2.x instead of any PIE 1.x.

For PHP itself it's exactly opposite: I consider the availability of `php:8` to be a mistake. PHP does not follow Semver and generally upgrading from one PHP branch to the next is not trivial, because there's always *some* breaking changes. So users should be encouraged to make an explicit choice when changing the release branch, allowing them to use `php:8` encourages the Semver myth.

Or as another example: HAProxy does not follow Semver either and instead simply treats version numbers as decimals. That means after x.9 comes (x+1).0. There was a `haproxy:1` under the false assumption that any HAProxy 2 would be a major release. With HAProxy 2.0 the “major version” variant was removed. There's only `haproxy:2.0` and `haproxy:2.0.0`. And nowadays also `haproxy:lts` to pin to the latest LTS version.

----------------

Any choice we make at this point will need to be maintained for an indefinite future, because once the versioning scheme cat is out of the bag it's hard to reverse it. By being conservative for now, we can first observe how using the Docker image works for for the users.

docs/usage.md
TimWolla: > These images are not meant to be used as PIE runtime, only as a interim carrier for easy `COPY --from` during custom Docker builds.

Yes, I understand that. I'm questioning the value of that. If as a user I need to think about the necessary steps to make PIE actually functional, then I might as well grab the pie.phar directly.

If use a Docker-optimized installation mechanism, then I expect it to simply work without any further action within a Docker context. 

> I am not sure if this should be PIE's responsibility to prepare runtime

It should not be PIE's responsibility, but the the responsibility of the distribution mechanism (i.e. the Docker image containing PIE). It can be as simple as a shell script like the following (untested):

```
#!/bin/bash

prepare_environment() {
	export DEBIAN_FRONTEND="noninteractive"
	savedAptMark="$(apt-mark showmanual)"
	buildDeps="
		libzip-dev
	"
	set -ex
	apt-get update
	apt-get install -y $buildDeps --no-install-recommends
	docker-php-ext-install -j "$(nproc)" zip
	apt-mark auto '.*' > /dev/null;
	apt-mark manual $savedAptMark;
	ldd "$(php -r 'echo ini_get("extension_dir");')"/*.so \
		| awk '/=>/ { so = $(NF-1); if (index(so, "/usr/local/") == 1) { next }; gsub("^/(usr/)?", "", so); print so }' \
		| sort -u \
		| xargs -r dpkg-query --search \
		| cut -d: -f1 \
		| sort -u \
		| xargs -r apt-mark manual
	apt-get purge -y --auto-remove -o APT::AutoRemove::RecommendsImportant=false
	rm -rf /var/lib/apt/lists/*
}

if ! pie --version >/dev/null 2>&1; then
    prepare_environment
fi

exec pie "$@"
```

Wirone: Downloading PHAR in Docker build may require installing `curl` or `wget`, while copying PHAR from Docker image can be done natively and is cached on builder level (no more downloading on subsequent builds). It's also more stable as Docker image and binary's name and won't change (regardless of changes in repo and release process), while URL to the PHAR can. I am not against providing installer, but it shouldn't be blocker here, since installer would require PHAR file anyway. Binary-only approach works well in Composer, I believe it has value here too ☺️.

docs/usage.md
```diff
@@ -29,6 +29,18 @@ system:
...
+
+PIE is published as binary-only Docker image, so you can install it easily during your Docker build:
+
+```Dockerfile
+COPY --from=ghcr.io/php/pie:latest /pie /usr/bin/pie
+```
+
+Instead of `latest` you can also use explicit versions like `x.y.z`, `x.y` or `x`, depending on stability level you want to achieve.
+
+If you want to verify signature of the PHAR, you need to additionally download `*.asc` file from [corresponding release](https://github.com/php/pie/releases).
```
asgrim, line 42: We have done away with the GPG signing now; this PR will need updating to use this for generating build provenance of this container image: https://docs.github.com/en/actions/security-for-github-actions/using-artifact-attestations/using-artifact-attestations-to-establish-provenance-for-builds#generating-build-provenance-for-container-images

.github/workflows/release.yml
```diff
@@ -50,3 +50,65 @@ jobs:
...
             pie.phar
             pie.phar.asc
+      - name: Store PHAR as artifact
+        uses: actions/upload-artifact@v4
+        with:
+          name: phar-binary
+          path: pie.phar
+
+  docker-binary-only-image:
+    needs: release-phar
```
asgrim, line 60: With a rebase, this would need to become:

```suggestion
    needs: build-phar
```

.github/workflows/release.yml
```diff
@@ -50,3 +50,65 @@ jobs:
...
+      packages: write
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Restore built PHAR
+        uses: actions/download-artifact@v4
+        with:
+          name: phar-binary
```
asgrim, line 76: After a rebase, this can become:

```suggestion
      - name: Fetch built PHAR from artifacts
        uses: actions/download-artifact@v4
        with:
          name: pie-${{ github.sha }}.phar
      - name: Verify the PHAR
        env:
          GH_TOKEN: ${{ github.token }}
        run: gh attestation verify pie.phar --repo ${{ github.repository }}
```

.github/workflows/release.yml
```diff
@@ -50,3 +50,65 @@ jobs:
...
+
+      - name: Build and push Docker image
+        uses: docker/build-push-action@v5
+        with:
+          context: .
+          platforms: linux/amd64,linux/arm64
+          file: Dockerfile
+          target: standalone-binary
+          build-args: |
+            PIE_VERSION=${{ github.ref_name }}
```
asgrim, line 111: This `PIE_VERSION` is unused now, can be dropped

docs/usage.md
```diff
@@ -29,6 +29,18 @@ system:
 sudo curl -L --output /usr/local/bin/pie https://github.com/php/pie/releases/latest/download/pie.phar && sudo chmod +x /usr/local/bin/pie
 ```
 
+### Docker installation
+
+PIE is published as binary-only Docker image, so you can install it easily during your Docker build:
+
+```Dockerfile
+COPY --from=ghcr.io/php/pie:latest /pie /usr/bin/pie
```
asgrim, line 37: I wonder if the image naming should be something like `ghcr.io/php/pie.phar:<tag>`, to help indicate that you can't simply run it (e.g. `docker run --rm ghrc.io/php/pie --version`) :thinking:

docs/usage.md
```diff
@@ -29,6 +29,18 @@ system:
 sudo curl -L --output /usr/local/bin/pie https://github.com/php/pie/releases/latest/download/pie.phar && sudo chmod +x /usr/local/bin/pie
 ```
 
+### Docker installation
+
+PIE is published as binary-only Docker image, so you can install it easily during your Docker build:
```
asgrim, line 34: We should probably explicitly note that this image doesn't have any dependencies; it's implied by "binary-only", but I think it would be clearer for people reading the docs.

asgrim (changes requested):

docs/usage.md
asgrim: > If as a user I need to think about the necessary steps to make PIE actually functional, then I might as well grab the pie.phar directly.

@TimWolla I understand what you're saying, but I think that's out of scope for this specific request. To alleviate this confusion, I think the image name we build here shouldn't be `ghcr.io/php/pie` but something more like `ghcr.io/php/pie-phar` or `ghcr.io/php/pie-bin`, to indicate that you can't actually "run" this image. Having a separate runnable Docker image with all the dependencies is something that could certainly be considered for addition, but I understand @Wirone's patch just provides that small benefit of `COPY --from=... /pie /usr/bin/pie` (I have personally used this technique a lot with Composer, so it is beneficial, IMO); I've also suggested documentation enhancements to make the fact you are responsible for installing the dependencies using this.

> Maybe we should add -bin suffix to make it explicit, just like it's done in Composer (then it's possible to provide fully usable images in the future without breaking naming convention).

@Wirone yes - I just caught this suggestion, and aligns with what I just wrote above :100:

Dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM scratch AS standalone-binary
+
+# @TODO change to --chmod=+x when https://github.com/moby/buildkit/pull/5380 is released
```
asgrim, line 3: Tracking this todo; looks like the PR is merged, I have no idea if it was released/usable yet though

Wirone: Yes it's merged and I even tested this at some point, but it wasn't working - seems like it's not released yet, hence the `@TODO`.

.github/workflows/release.yml
Wirone: Applied during interactive rebase.

.github/workflows/release.yml
Wirone: Removed during interactive rebase.

.github/workflows/build-phar.yml
```diff
@@ -61,3 +61,8 @@ jobs:
         with:
           name: pie-${{ github.sha }}.phar
           path: pie.phar
+      - name: Store PHAR as artifact
```
asgrim: This is not necessary, the step before already stores the PHAR

.github/workflows/release.yml
asgrim: This still needs to be added; please use the existing artifact (`pie-${{ github.sha }}.phar`), and add the `attestation verify` step

docs/usage.md
asgrim: That's fine IMO, thank you

docs/usage.md
asgrim: Thanks, looks good

.github/workflows/build-phar.yml
asgrim: NM, was just updated

.github/workflows/release.yml
asgrim: NM was just updated

.github/workflows/build-phar.yml
Wirone: Yeah, my bad, I did not change restoring artifact during rebase and it failed, I thought storing was missing. Fixed already.

.github/workflows/release.yml
Wirone: Done in [e0c258d](https://github.com/php/pie/pull/122/commits/e0c258d53b3563af89687967013260de6aafff99).

docs/usage.md
Wirone: Attestation step added, [here's the result](https://github.com/Wirone/php-pie/attestations/3509753).

Wirone:
@asgrim Docker build attestation from [latest build](https://github.com/Wirone/php-pie/actions/runs/12083632706/job/33697195403) can be found [here](https://github.com/Wirone/php-pie/attestations/3509753).

Dockerfile
Wirone: It was released under [experimental `1.11.0-labs` tag](https://github.com/moby/buildkit/releases/tag/dockerfile/1.11.0-labs) and can be used as opt-in by adding `# syntax=docker.io/docker/dockerfile:1.11.0-labs` to the Dockerfile. Do we want to add it, or just update it when it's released in stable version?

Wirone:
@asgrim Is there any actionable thing on my side? I believe [this](https://github.com/php/pie/pull/122#discussion_r1863364125) can wait for non-labs release, as this is only a cosmetic change.

TimWolla (commented):
I'm still concerned about https://github.com/php/pie/pull/122/files#r1854054535, given that we only make the Phar available as “latest” and as a specific version, not by any “major” / “minor” wildcards.

The workflow steps look fine to me, except for the few remarks I've made.

.github/workflows/release.yml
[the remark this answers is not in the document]
Wirone: PHP or HAProxy versioning aside (out of the scope), what's the actual problem with availability of fine-tuned tags for PIE's Docker images? Is it against the versioning policy, in this case SemVer, that I believe is being used in this project? My perspective is:

- `pie:X-bin` is for people who always want to use latest semver-compatible version
- `pie:X.Y-bin` is for people who does not want to deprecation warnings pop up randomly during build (deprecations are supposed to be in minor versions)
- `pie:X.Y.Z-bin` is for people who want to pin-point exact version to have stable and reproducible builds

I believe all those strategies are valid, and having all the tags make it possible to choose the most fitting one.

`latest-bin` (or just `bin`, even though I am not in favour of this) on the other hand is for YDD (YOLO Driven Development) or experiments, it's also good to have it, since it does not cost anything 🙂.

.github/workflows/release.yml
```diff
@@ -6,7 +6,10 @@ on:
       - published
 
 permissions:
+  attestations: write
   contents: read
+  id-token: write
+  packages: write
```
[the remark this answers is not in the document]
Wirone: I will move the permissions to specific job, but are descriptions really needed? Permissions are [documented and explained](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/controlling-permissions-for-github_token), I don't see a value in providing the same info every time permission is used within multiple jobs. In most cases it can be inferred from the job context.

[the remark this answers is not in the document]
Wirone: I added descriptions anyway, but I am not sure if I am right about `id-token:write` 😅.

docs/usage.md
```diff
@@ -27,6 +27,19 @@ system:
...
+
+```Dockerfile
+COPY --from=ghcr.io/php/pie:latest-bin /pie /usr/bin/pie
```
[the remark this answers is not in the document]
TimWolla: see https://github.com/docker-library/official-images?tab=readme-ov-file#tags-and-aliases

.github/workflows/release.yml
```diff
@@ -38,3 +38,77 @@ jobs:
...
+    needs: build-phar
+    name: Docker binary-only image
+    runs-on: ubuntu-latest
+    if: ${{ startsWith(github.ref, 'refs/tags/') }}
+
+    permissions:
+      # attestations:write is required to publish provenance attestations for the built Docker image
+      attestations: write
+      # id-token:write is required for logging in to the Docker registry
+      id-token: write
```
TimWolla, line 52:
```suggestion
      # id-token:write is required for build provenance attestation.
      id-token: write
      # attestations:write is required for build provenance attestation.
      attestations: write
```

Taken from: https://github.com/php/pie/blob/main/.github/workflows/build-phar.yml

docs/usage.md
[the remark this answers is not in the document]
Wirone: OK, that makes me change my mind, and `pie:bin` should be enough as the newest binary-only image. I'll push changes soon.

.github/workflows/release.yml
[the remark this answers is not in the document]
TimWolla: > but I am not sure if I am right about `id-token:write` 😅.

Which is a good argument in favor of the descriptions. That makes it easy to check whether a given permission is still required.

`id-token:write` is also required for attestation. It's the permission that allows to turn the `GITHUB_TOKEN` into a JWT, which allows to prove to *external* consumers that the call is made from within GitHub Actions.

One example use case would be an AWS policy that allows these JWTs to access specific resources (e.g. to create an EC2 instance for deployment), without needing to manually store secrets.

.github/workflows/release.yml
Wirone: I've applied changes to descriptions, but kept permissions ordered.

docs/usage.md
```diff
@@ -27,6 +27,19 @@ system:
...
+```
+
+Instead of `latest` you can also use explicit versions like `x.y.z-bin`, `x.y-bin` or `x-bin`, depending on stability level you want to achieve.
```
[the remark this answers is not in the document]
Wirone: Changed to `bin`.

.github/workflows/release.yml
```diff
@@ -38,3 +38,76 @@ jobs:
...
+        uses: docker/metadata-action@v5
+        with:
+          flavor: |
+            latest=false
+          images: ghcr.io/${{ github.repository }}
+          tags: |
+            type=raw,value=bin
+            type=semver,pattern={{version}}-bin
+            type=semver,pattern={{major}}.{{minor}}-bin
+            type=semver,pattern={{major}}-bin
```
asgrim, line 94: Thank you for all the work you're doing on this @Wirone - I think this is nearly close to being ready. At this stage, given the early stage of releases, I would like to reduce these tags down to:

```suggestion
          tags: |
            type=raw,value=bin
            type=semver,pattern={{version}}-bin
```

Once we release `1.0.0` we can re-open a fresh discussion about the `:{{major}}.{{minor}}-bin` and `:{{major}}-bin` tags; but at this stage, this is unnecessary, and since semver prior to `1.0.0` works differently anyway (i.e. each minor may contain a BC break), then this currently may have unintended consequences.

I had considered _only_ keeping `:bin`, but I think keeping `:{{version}}-bin` has the benefit of people being able to pin to a specific version should there be a compatibility issue or bug introduced in a newer release.

Dockerfile
asgrim: There is a newer release 1.12 now, is this possible to use with a stable version?

Wirone: 1.12 does not contain this. I just tested labs release

```Dockerfile
# syntax=docker.io/docker/dockerfile:1.11.0-labs

FROM scratch AS standalone-binary

COPY --chmod=+x pie.phar /pie
```

But it does not work somehow 🤷:

```
$  docker build --progress=plain --no-cache -t pie-local .
#0 building with "orbstack" instance using docker driver

#1 [internal] load build definition from Dockerfile
#1 transferring dockerfile: 152B done
#1 DONE 0.0s

#2 resolve image config for docker-image://docker.io/docker/dockerfile:1.11.0-labs
#2 ...

#3 [auth] docker/dockerfile:pull token for registry-1.docker.io
#3 DONE 0.0s

#2 resolve image config for docker-image://docker.io/docker/dockerfile:1.11.0-labs
#2 DONE 1.9s

#1 [internal] load build definition from Dockerfile
#1 transferring dockerfile: 152B done
#1 DONE 0.0s

#4 docker-image://docker.io/docker/dockerfile:1.11.0-labs@sha256:51bca3eb9dfe8ad06ee1b137d3fa44f3cef79200d5b0010b580da63d128609e4
#4 CACHED

#5 [internal] load build definition from Dockerfile
#5 DONE 0.0s

#6 [internal] load .dockerignore
#6 transferring context: 2B done
#6 DONE 0.0s
Dockerfile:5
--------------------
   3 |     FROM scratch AS standalone-binary
   4 |     
   5 | >>> COPY --chmod=+x pie.phar /pie
   6 |     
--------------------
ERROR: failed to solve: invalid chmod parameter: '+x'. it should be octal string and between 0 and 07777
```

Let me know if you want to keep `@TODO` or just stick with currently used `--chmod=0755` (remove the comment), which works correctly.

.github/workflows/release.yml
Wirone: Done in [460f0c3](https://github.com/php/pie/pull/122/commits/460f0c380159bf8040a45d42b946279dfba64bf6), let me know if the `@TODO` is OK for you, to track this before v1.0.

Wirone: @asgrim FYI the link `https://ghcr.io/php/pie` used in usage docs currently returns 404 because there are no built images in the registry. Please see `https://ghcr.io/wirone/php-pie` (my fork's short URL) to see it redirects to working registry, so it will also work here after first release.

TimWolla (approved):
LGTM now. Thanks for your patience with my reviews :smile:

Wirone:
@TimWolla after @keradus' reviews nothing seems to be annoying 😂. But seriously, thanks for the great feedback, I've learned new things during the process so for me this was added value. I like [good code reviews](https://blog.codito.dev/2022/07/code-review-good-practices/) 👍.

Dockerfile
asgrim: OK lets leave as-is for now

asgrim (approved):
Thank you @Wirone for your patience, this LGTM now :+1: got there in the end :grin: nice work, and thanks @TimWolla for your great feedback too!

glensc:
1.x released, so adding the versioned "-bin" tags make sense now:
- https://github.com/php/pie/pull/614

PR #127, opened 2024-11-22
Configure fine-grained GitHub Action permissions

TimWolla:
I was not able to test the workflows requiring the additional permissions, because I obviously do not have a signing key set up. I'm reasonably confident that these changes are correct though.

@asgrim I'm not sure if you are able to make the necessary changes to the repository configuration. You might need to request assistance by a organization owner.

-----------------

This change configures GitHub Actions to only allow reading the repository contents within the top-level workflow-wide permissions section. Extended permissions are then granted to individual jobs to make permissions “secure by default” even when additional jobs are added to an existing workflow.

As a follow-up to this PR, the “Workflow Permissions” in the repository settings should be reconfigured to “Read repository contents and packages permissions” instead of “Read and write permissions” to also make additional workflows secure by default, even when a permissions section is not explicitly configured.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -4,6 +4,9 @@ on:
   pull_request:
   push:
 
+permissions:
+  contents: read
+
 jobs:
   unit-tests:
     runs-on: ${{ matrix.operating-system }}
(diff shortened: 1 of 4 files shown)


asgrim (approved):

asgrim:
The **Workflow permissions** has been updated to **Read repository contents and packages permissions** :heavy_check_mark: 

Thanks @TimWolla :+1:

PR #128, opened 2024-11-22
Add build provenance attestation to the release workflow

TimWolla:
This PR builds upon #127 and includes that one as a subset. I will rebase it after #127 is merged.

-------------------

This attestation proves that a given pie.phar was created by the release.yml workflow running for a specific commit.

Verification is possible using the `gh` CLI as follows:

    gh attestation verify --owner php pie.phar

For the test in my fork, the output looked like this:

    $ gh attestation verify --owner TimWolla pie.phar
    Loaded digest sha256:0dc7f9c88341fc1beac5746658e4a859a1abe8b3f6a62aa9f10c9ec39ff42b53 for file://pie.phar
    Loaded 2 attestations from GitHub API
    ✓ Verification succeeded!

    sha256:0dc7f9c88341fc1beac5746658e4a859a1abe8b3f6a62aa9f10c9ec39ff42b53 was attested by:
    REPO              PREDICATE_TYPE                  WORKFLOW
    TimWolla/php-pie  https://slsa.dev/provenance/v1  .github/workflows/release.yml@refs/tags/release-action-build-provenance-1
    TimWolla/php-pie  https://slsa.dev/provenance/v1  .github/workflows/release.yml@refs/heads/release-action-build-provenance

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -20,6 +20,10 @@ jobs:
     permissions:
       # contents:write is required to upload the binaries to the release.
       contents: write
+      # id-token:write is required for build provenance attestation.
+      id-token: write
+      # attestations:write is required for build provenance attestation.
+      attestations: write
     steps:
       - name: Setup PHP
         uses: shivammathur/setup-php@v2
@@ -49,6 +53,10 @@ jobs:
               --detach-sign \
               --output pie.phar.asc \
               pie.phar
+      - name: Generate build provenance attestation
+        uses: actions/attest-build-provenance@v1
+        with:
+          subject-path: '${{ github.workspace }}/pie.phar'
       - name: Upload binaries to release
         uses: softprops/action-gh-release@v2
         if: ${{startsWith(github.ref, 'refs/tags/') }}


.github/workflows/release.yml
```diff
@@ -43,6 +53,10 @@ jobs:
               --detach-sign \
               --output pie.phar.asc \
               pie.phar
+      - name: Generate build provenance attestation
+        uses: actions/attest-build-provenance@v1
```
asgrim, line 57: If I'm understanding correctly, this should entirely negate the need to "manually" sign the phar, right? Or is there something I'm missing? I'm wondering if we can rip out the GPG signing stuff, and update the documentation for the verification to use the `gh attestation verify ...` command.

TimWolla: Yes, this is my understanding. The only possible drawback would be that PGP is much more easily available compared to GitHub CLI. Debian for example bundles GitHub CLI only since the newest stable version and only in version 2.23, which does not yet include the `gh attestation` sub-command.

I'm also just noticing that the attestation functionality also has support for [attesting the SBOM](https://docs.github.com/en/actions/security-for-github-actions/using-artifact-attestations/using-artifact-attestations-to-establish-provenance-for-builds#generating-an-attestation-for-a-software-bill-of-materials-sbom), which might be an interesting follow-up.

TimWolla:
Rebased now.

TimWolla:
Test failure is unrelated. Might be necessary to implicitly pass `$GITHUB_TOKEN` to PIE for API authentication?

.github/workflows/release.yml
asgrim: I think needing `gh` tool is an acceptable tradeoff. The attestations I believe are accessible via an API call anyway, which `gh` wraps, so in theory another tool could be used to fetch that.

I'll do this in a separate PR, as I want to consolidate the release pipelines too.

asgrim: #132

asgrim (approved):
Thanks @TimWolla !

PR #131, opened 2024-11-25
chore: move the obtaining the version method to a separate class

godruoyi:


src/Util/PipVersion.php
```diff
@@ -0,0 +1,41 @@
...
+
+declare(strict_types=1);
+
+namespace Php\Pie\Util;
+
+use Composer\InstalledVersions;
+
+use function class_exists;
+
+class PipVersion
```
asgrim, line 11: Typo in class and filename, should be `PieVersion`. Should also be `final` as this should not be overridden:

```suggestion
final class PieVersion
```

asgrim (changes requested):

src/Util/PipVersion.php
godruoyi: nice catch, updated, @asgrim PTAL,

asgrim (approved):

asgrim:
Thanks @godruoyi !

src/Util/PieVersion.php
```diff
@@ -0,0 +1,69 @@
...
+use Composer\InstalledVersions;
+
+use function class_exists;
+
+/** @internal This is not public API for PIE, so should not be depended upon unless you accept the risk of BC breaks */
+final class PieVersion
+{
+    /**
+     * Note: magic constant that causes Symfony Console to not display a version
+     * {@see Application::getLongVersion()}
```
TimWolla, line 16: This is a dead reference, because `Application` is not imported here.

asgrim: Hah, good spot - indeed it's not fully qualified, but PhpStorm doesn't seem to mind, I'd actually checked I could click through, so missed it wasn't fully qualified

Screenshot from 2024-11-26 09-24-37

asgrim: @godruoyi I actually added a comment to the class in your PR:

https://github.com/php/pie/blob/75dbc227ce6445fab7b40bc8f99fdd4ca76ebbcc/src/Util/PieVersion.php#L29-L41

Issue #140, opened 2024-11-27
PHP 8.4 compatibility


ruudk:
It seems that the latest 0.2.0 as well as the nightly is not fully compatible with PHP 8.4 as it produces deprecation warnings:

```command
$ php -v
PHP 8.4.1 (cli) (built: Nov 20 2024 09:48:35) (NTS)
Copyright (c) The PHP Group
Zend Engine v4.4.1, Copyright (c) Zend Technologies
    with Zend OPcache v8.4.1, Copyright (c), by Zend Technologies
```

```command
$ php pie-nightly.phar --version
Deprecated: Illuminate\Container\Container::beforeResolving(): Implicitly marking parameter $callback as nullable is deprecated, the explicit nullable type must be used instead in phar:///Volumes/CS/www/cosmos/pie-nightly.phar/vendor/illuminate/container/Container.php on line 1151

Deprecated: Illuminate\Container\Container::resolving(): Implicitly marking parameter $callback as nullable is deprecated, the explicit nullable type must be used instead in phar:///Volumes/CS/www/cosmos/pie-nightly.phar/vendor/illuminate/container/Container.php on line 1171

Deprecated: Illuminate\Container\Container::afterResolving(): Implicitly marking parameter $callback as nullable is deprecated, the explicit nullable type must be used instead in phar:///Volumes/CS/www/cosmos/pie-nightly.phar/vendor/illuminate/container/Container.php on line 1191

Deprecated: Illuminate\Container\Container::setInstance(): Implicitly marking parameter $container as nullable is deprecated, the explicit nullable type must be used instead in phar:///Volumes/CS/www/cosmos/pie-nightly.phar/vendor/illuminate/container/Container.php on line 1430

Deprecated: Illuminate\Contracts\Container\Container::beforeResolving(): Implicitly marking parameter $callback as nullable is deprecated, the explicit nullable type must be used instead in phar:///Volumes/CS/www/cosmos/pie-nightly.phar/vendor/illuminate/contracts/Container/Container.php on line 200

Deprecated: Illuminate\Contracts\Container\Container::resolving(): Implicitly marking parameter $callback as nullable is deprecated, the explicit nullable type must be used instead in phar:///Volumes/CS/www/cosmos/pie-nightly.phar/vendor/illuminate/contracts/Container/Container.php on line 209

[...]

PR #141, opened 2024-11-27
Update illuminate/container and illuminate/contracts

TimWolla:
This fixes the nullable parameter deprecation for PHP 8.4 and thus fixes php/pie#140.


ruudk:
Just tested the phar and can confirm that it solves the problem! Great!

asgrim (approved):

asgrim:
Thanks @TimWolla !

PR #143, opened 2024-11-28
feat: Add PHP version validation functionality

tal7aouy:
This PR adds version validation capabilities to the `PhpBinaryPath` class, allowing users to verify minimum PHP version requirements. This is particularly useful for ensuring compatibility with different PHP installations.

## Changes
- Added `assertMinimumVersion()` method to `PhpBinaryPath`
- Added comprehensive version validation tests
- Improved version handling and validation
- Added version caching for better performance

## Example Usage
```php
$php = PhpBinaryPath::fromCurrentProcess();

// Will throw exception if current PHP version is less than specified
$php->assertMinimumVersion('8.1.0');
```

### DIFF
--- a/src/Platform/TargetPhp/PhpBinaryPath.php
+++ b/src/Platform/TargetPhp/PhpBinaryPath.php
@@ -273,4 +273,23 @@ PHP,
 
         return new self($phpExecutable, null);
     }
+
+    /**
+     * @param non-empty-string $minimumVersion
+     * @throws RuntimeException if version requirement not met
+     */
+    public function assertMinimumVersion(string $minimumVersion): void
+    {
+        $versionParser = new VersionParser();
+        $normalized = $versionParser->normalize($this->version());
+        $normalizedMin = $versionParser->normalize($minimumVersion);
+        
+        if (version_compare($normalized, $normalizedMin, '<')) {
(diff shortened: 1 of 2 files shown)


src/Platform/TargetPhp/PhpBinaryPath.php
```diff
@@ -273,4 +273,23 @@ public static function fromCurrentProcess(): self
 
         return new self($phpExecutable, null);
     }
+
+    /**
+     * @param non-empty-string $minimumVersion
+     * @throws RuntimeException if version requirement not met
+     */
+    public function assertMinimumVersion(string $minimumVersion): void
```
asgrim, line 281: I don't think this is really needed, since the dependency resolution is already based on the target PHP version... What is the use case here? There aren't actually any usages of this new method...?

tal7aouy: You're absolutely right. I was thinking about runtime version validation but didn't consider that Composer's dependency resolution already handles this effectively through the `composer.json` requirements:

```json
{
    "require": {
        "php": ">=8.1"
    }
}
```

Since there's no immediate use case requiring runtime version validation beyond what Composer provides, I'll close this PR. Thank you for helping me avoid unnecessary code complexity!

PR #148, opened 2024-12-04
add remi's supported ext

remicollet:


remicollet:
Notice:  could be nice if this list was ordered

asgrim:
Thanks @remicollet :) the order is "the order on the PECL website" :laughing:  which I think is ordered by number of downloads or some statistics

PR #149, opened 2024-12-05
Add missing dependencies to composer.json

TimWolla:
Code from these packages is explicitly referenced in PIE itself, but previously these packages were only required transitively through other dependencies.


asgrim (approved):
LGTM :+1:

PR #155, opened 2024-12-12
Automatically setup php.ini entries

### LINKED ISSUE #20: Detect and automatically set up `php.ini` entry for installation

asgrim:
Fixes #20 

 - [x] ensure auto setup can be disabled with `--skip-enable-extension` flag
 - [x] set up single `php.ini`
 - [x] set up `00-extname.ini` in additional php files path
 - [x] set up with `docker-php-ext-enable` for Docker users
 - [x] set up with `phpenmod` for `ppa:ondrej/php` users


remicollet:
IMHO, tasks wording should be 

*  set up with phpenmod for ppa:ondrej/php users => set up for DEB packages users (AFAIK, nothing specific to ondrej ppa)
* set up for Remi repo users (need to check how this is done) => set up for RPM packages users (at least Fedora / RHEL / centOS and clones, don't know for other distros such as suse / mageia....)

At least for RPM, the numerical prefix is NOT related to "**priority**" but really to "**load order**". This prefix was introduced to ensure proper load order, according to declared dependencies, to ensure all symbols are resolved (PHP is build using the `--enable-rtld-now` option).

I think distribution package users, should not mix with other installation methods.

If really needed (not available as package), any number greater what packages use should be fine

For RPM, number are described in [Guidelines](https://docs.fedoraproject.org/en-US/packaging-guidelines/PHP/#c-pecl-config-file)

oerdnj:
20 is a baseline, but if extensions stack upon each other, it usually goes in increments of 5, so memcached is something like 30.

phpenmod works with `;priority=nn` comment in the ini file.

asgrim:
>     * set up with phpenmod for ppa:ondrej/php users => set up for DEB packages users (AFAIK, nothing specific to ondrej ppa)

the `ppa:ondrej/php` is special, because it uses `phpenmod` to enable extensions. My plan is to check if `phpenmod` exists in the path, and to try to enable the extension using that, instead of the standard approach of enabling the extension.
 
>     * set up for Remi repo users (need to check how this is done) => set up for RPM packages users (at least Fedora / RHEL / centOS and clones, don't know for other distros such as suse / mageia....)

I suppose my question is: is there anything similar to `phpenmod` for the Remi repo? If not, then there is nothing to do here IMO.

> At least for RPM, the numerical prefix is NOT related to "**priority**" but really to "**load order**". This prefix was introduced to ensure proper load order, according to declared dependencies, to ensure all symbols are resolved (PHP is build using the `--enable-rtld-now` option).

The configuration `priority` that PIE-compatible extension maintainers can optionally specify will use that. If not specified by the extension author, the `priority` used is 80. The `priority` is only used where the additional INI file directory is specified (shown as `Scan this dir for additional .ini files` in the info).

PR #158, opened 2024-12-18
Trim warnings and deprecations from output when invoking PHP in `PhpBinaryPath`

### LINKED ISSUE #146: PHP startup warnings in target PHP result in `InvalidPhpBinaryPath` exception
How to reproduce:

 - modify your `php.ini` to load an extension that does not exist, e.g. `extension=blah`
 - try to run any PIE operation that uses that target PHP to whom the `php.ini` belongs to

What should happen:

 - `\Php\Pie\Platform\TargetPhp\PhpBinaryPath::assertValidLookingPhpBinary` should not throw an exception

What actually happens:

 - the exception is raised in `\Php\Pie\Platform\TargetPhp\Exception\InvalidPhpBinaryPath::fromInvalidPhpBinary`

Additional information:

The output from the "valid-looking PHP" check is different. We expect simply the string `PHP`, but we see: 

```
Warning: PHP Startup: Unable to load dynamic library 'blah' (tried: <redacted bunch of stuff it tried>) in Unknown on line 0
PHP
```

Hints; would be worth checking if the warning goes to `stderr`; this may help us easily filter out the warning (rather than doing some kind of string-based filtering) as then we can just check `stderr` output is exactly the string `"PHP"` in the check.

alexandre-daubois:
Fix #146 

Adds a way to trim `Deprecated:` and `Warning:` from output. This is not only limited to `assertValidLookingPhpBinary`, but to all PHP subprocess invocation.

Unfortunately, warning messages are sent to `stdout`, not `stderr`.

The cleaning only happens during some setup/validation step of the current env, and the deprecations and warnings are still being displayed to the user when compiling the extension. This way, there's no "sneaky hiding" of error:

image

I added a parameter to the `Process::run()` method, but if we don't want to add it, the output could also be directly cleaned in the `PhpBinaryPath`. Indeed, I'm not _100%_ happy of this parameter, which is really "PHP-subprocess-related" in a class that is not limited to PHP subprocesses.

### DIFF
--- a/features/install-extensions.feature
+++ b/features/install-extensions.feature
@@ -22,6 +22,11 @@ Feature: Extensions can be installed with Behat
     When I run a command to build an extension
     Then the extension should have been built
 
+  Example: An extension can be built with warnings at PHP startup
+    Given I use the "-d extension=invalid_extension.so" PHP arguments
+    When I run a command to build an extension
+    Then the extension should have been built
+
   Example: An extension can be built with configure options
     When I run a command to build an extension with configure options
     Then the extension should have been built with options
--- a/features/install-extensions.feature
+++ b/features/install-extensions.feature
@@ -23,7 +23,7 @@ Feature: Extensions can be installed with Behat
     Then the extension should have been built
 
   Example: An extension can be built with warnings at PHP startup
-    Given I use the "-d extension=invalid_extension.so" PHP arguments
+    Given I have an invalid extension installed
     When I run a command to build an extension
     Then the extension should have been built
 
--- a/src/Platform/TargetPhp/PhpBinaryPath.php
+++ b/src/Platform/TargetPhp/PhpBinaryPath.php
@@ -20,6 +20,7 @@ use function assert;
 use function dirname;
 use function explode;
 use function file_exists;
+use function implode;
 use function is_dir;
 use function is_executable;
 use function preg_match;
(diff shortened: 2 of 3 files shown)


src/Util/Process.php
```diff
@@ -36,9 +39,31 @@ public static function run(
...
             ->getOutput());
+
+        if ($cleanWarningsAndDeprecations) {
+            return self::cleanLanguageWarningAndDeprecations($output);
+        }
+
+        return $output;
+    }
+
+    private static function cleanLanguageWarningAndDeprecations(string $testOutput): string
```
asgrim, line 55: Hmm; I don't think this should be part of `Process`, it should be in `PhpBinaryPath`, since it is specifically related to PHP; the `Process` we are using for other non-PHP executables too, so it doesn't make sense in those cases IMO. Could this be moved to `PhpBinaryPath` please? :)

src/Util/Process.php
```diff
@@ -36,9 +39,31 @@ public static function run(
...
+    {
+        $testOutput = explode("\n", $testOutput);
```
asgrim, line 57: This might need to be `PHP_EOL`; I don't think we have any tests to reproduce this at the moment, so I don't actually know if Windows would use `\r\n` in the process output. Worth verifying with a test IMO!

asgrim:
> Unfortunately, warning messages are sent to `stdout`, not `stderr`.

Dang! That's frustrating, but yeah the approach you've taken is fine given that :+1: thank you @alexandre-daubois :)

src/Util/Process.php
alexandre-daubois: Yes I wasn't sure where to put this, but I guess calling the cleaning method in `PhpBinaryPath` is better. That's updated 👍

src/Util/Process.php
alexandre-daubois: Added a behavior test that tries to load an invalid extension. Works well 👍

asgrim (approved):

asgrim:
Nice one, thanks @alexandre-daubois !

Issue #89, opened 2024-11-14
Interactive GitHub OAuth token request fails


TimWolla:
Using #83:

I was attempting to download a package from a custom GitHub VCS in a fresh Docker container without any composer credentials set up. The GitHub API requests were rejected by GitHub's API, with PIE emitting an informational message that credentials were required. However it crashed while actually requesting the credentials:

```
root@26098011df0d:/pwd# php bin/pie download -v redacted/redacted:dev-pie
You are running PHP 8.3.13
Target PHP installation: 8.3.13 nts, on Linux/OSX/etc x86_64 (from /usr/local/bin/php)
Reading composer.json of https://github.com/redacted/redacted.git (v0.2.0)GitHub API limit (60 calls/hr) is exhausted, could not fetch https://api.github.com/repos/redacted/redacted/contents/composer.json?ref=redacted. Create a GitHub OAuth token to go over the API rate limit. You can also wait until 2024-11-14 13:45:56 for the rate limit to reset.
When working with _public_ GitHub repositories only, head here to retrieve a token:
https://github.com/settings/tokens/new?scopes=&description=Composer+on+26098011df0d+2024-11-14+1246
This token will have read-only permission for public information only.
When you need to access _private_ GitHub repositories as well, go to:
https://github.com/settings/tokens/new?scopes=repo&description=Composer+on+26098011df0d+2024-11-14+1246
Note that such tokens have broad read/write permissions on your behalf, even if not needed by Composer.
Tokens will be stored in plain text in "/root/.composer/auth.json" for future use by Composer.
For additional information, check https://getcomposer.org/doc/articles/authentication-for-private-packages.md#github-oauth
Reading composer.json of https://github.com/redacted/redacted.git (v0.1.0)GitHub API limit (60 calls/hr) is exhausted, could not fetch https://api.github.com/repos/redacted/redacted/contents/composer.json?ref=redacted. Create a GitHub OAuth token to go over the API rate limit. You can also wait until 2024-11-14 13:45:56 for the rate limit to reset.
When working with _public_ GitHub repositories only, head here to retrieve a token:
https://github.com/settings/tokens/new?scopes=&description=Composer+on+26098011df0d+2024-11-14+1246
This token will have read-only permission for public information only.
When you need to access _private_ GitHub repositories as well, go to:
https://github.com/settings/tokens/new?scopes=repo&description=Composer+on+26098011df0d+2024-11-14+1246
[...]

PR #163, opened 2024-12-28
Allow composer to ask for credentials

Bellangelo:
Fixes https://github.com/php/pie/issues/89

I have created the value object `MinimalHelperSet` so we can always ensure that the `question` option exist.


asgrim (approved):

asgrim:
Thanks @Bellangelo !

PR #172, opened 2025-01-07
Improve wording when --skip-enable-extension

ruudk:
When `--skip-enable-extension` is used, it's done on purpose.

No need to uppercase not.
Explain how the extension can be enabled.

### DIFF
--- a/src/Installing/SetupIniFile.php
+++ b/src/Installing/SetupIniFile.php
@@ -42,9 +42,9 @@ class SetupIniFile
                 $output->writeln('Automatic extension enabling was skipped.', OutputInterface::VERBOSITY_VERBOSE);
             }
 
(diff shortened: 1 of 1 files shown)


src/Installing/SetupIniFile.php
```diff
@@ -42,9 +42,9 @@ public function __invoke(
                 $output->writeln('Automatic extension enabling was skipped.', OutputInterface::VERBOSITY_VERBOSE);
             }
 
-            $output->writeln('<comment>⚠️  Extension has NOT been automatically enabled.</comment>');
+            $output->writeln('<comment>⚠️  Extension is not enabled.</comment>');
```
asgrim, line 45: This messaging is not only emitted when `--skip-enable-extension` is used; it is also when we failed to automatically enable the extension. The messaging on the line above (`Automatic extension enabling was skipped.`) provides the context, although only with verbose mode.

Issue #169, opened 2025-01-02
Pass the GH token to tests


asgrim:
Occasionally, tests will fail because we are using GH API without an auth token. I suspect making sure the `GITHUB_TOKEN` in CI will eliminate the issue.

https://github.com/php/pie/actions/runs/12572758302/job/35045103222?pr=166

```
There was 1 error:

1) Php\PieIntegrationTest\Downloading\GithubPackageReleaseAssetsTest::testDeterminingReleaseAssetUrlForWindows
Composer\Downloader\TransportException: The "https://api.github.com/repos/asgrim/example-pie-extension/releases/tags/2.0.2" file could not be downloaded (HTTP/2 403 ):
{"message":"API rate limit exceeded for 13.91.71.16. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)","documentation_url":"htt...

D:\a\pie\pie\src\Downloading\GithubPackageReleaseAssets.php:80
D:\a\pie\pie\src\Downloading\GithubPackageReleaseAssets.php:38
D:\a\pie\pie\test\integration\Downloading\GithubPackageReleaseAssetsTest.php:68
```

asgrim:
Didn't work :facepalm: https://github.com/php/pie/actions/runs/12686877443/job/35360110779

Bellangelo:
Hi @asgrim, I think I figured out why your fix didn't work originally. Please check: https://github.com/php/pie/pull/174

PR #174, opened 2025-01-10
Fix access to the GITHUB_TOKEN

Bellangelo:
Closes https://github.com/php/pie/issues/169

The main issue was that while this [test](https://github.com/php/pie/pull/171/files#diff-ee4139668bc93757e55104b542b5de1e68322ed1fcf463b523835f5aa781dd82R65-R74) was trying to use the `GITHUB_TOKEN` environment variable, the variable was scoped to be used in a specific command only. I moved the env to the top so it can be accessed by all the commands in the step.

Furthermore, this line `- run: composer config --global github-oauth.github.com ${{ secrets.GITHUB_TOKEN }}` is removed because this is handled by the `setup-php` action. For references please check https://github.com/shivammathur/setup-php/blob/main/src/scripts/tools/add_tools.sh#L64 and https://github.com/shivammathur/setup-php?tab=readme-ov-file#wrench-tools-support

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -10,8 +10,6 @@ permissions:
 jobs:
   unit-tests:
     runs-on: ${{ matrix.operating-system }}
-    env:
-      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
     strategy:
       matrix:
         operating-system:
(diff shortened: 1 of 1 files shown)


asgrim (approved):

asgrim:
Thanks @Bellangelo !

PR #178, opened 2025-01-27
Remove `php-standard-library/psalm-plugin`

TimWolla:
The `azjezz/psl` dependency was removed in php/pie#42, making this dependency obsolete.


andrew-demb:
You need to update `psalm.xml.dist` too

TimWolla:
> You need to update `psalm.xml.dist` too

Thanks. I only searched for the full plugin name and didn't realize the namespace was different / the plugin was referenced by class-name.

asgrim (approved):
Thanks!

PR #181, opened 2025-01-29
Update to Psalm 6.x

TimWolla:
Opted to baseline the newly reported errors for now. Properly “fixing” those will likely result in tons of `assert()` calls being added.


GromNaN:
You should set `ignoreInternalFunctionFalseReturn` to `true`. The functions that return `false` in case of failure also trigger a warning, that is converted to an `Error` by the error handler (provided by Symfony Console or PHPUnit). So they never return in case of failure.

GromNaN (approved):

asgrim (approved):

asgrim:
Thank you!

PR #187, opened 2025-02-04
Allow extensions to override download URLs

### LINKED ISSUE #39: Investigate if git submodules can be included
At the moment, PIE downloads the ZIP archive, which doesn't include git submodules. [mongodb](https://github.com/mongodb/mongo-php-driver/blob/v1.x/.gitattributes) does use submodules for the C libs to be included.

The `.gitmodules` file is present, however, the revision information is kept in `.git` which isn't included in the Zip archive.

asgrim:
Fixes #39


test/unit/ComposerIntegration/Listeners/OverrideDownloadUrlInstallListenerTest.php
```diff
@@ -85,6 +85,21 @@ public function testEventListenerRegistration(): void
         );
     }
 
+    public function testNonInstallOperationsAreIgnored(): void
+    {
+        self::markTestIncomplete('todo'); // @todo
```
asgrim, line 90: Tests to be completed

test/unit/Downloading/DownloadUrlMethodTest.php
```diff
@@ -0,0 +1,28 @@
...
+use Php\Pie\Downloading\DownloadUrlMethod;
+use PHPUnit\Framework\Attributes\CoversClass;
+use PHPUnit\Framework\TestCase;
+
+#[CoversClass(DownloadUrlMethod::class)]
+final class DownloadUrlMethodTest extends TestCase
+{
+    public function testWindowsPackages(): void
+    {
+        self::markTestIncomplete('todo'); // @todo
```
asgrim, line 16: Tests to be completed

test/unit/Platform/PrePackagedSourceAssetNameTest.php
```diff
@@ -0,0 +1,18 @@
...
+use Php\Pie\Platform\PrePackagedSourceAssetName;
+use PHPUnit\Framework\Attributes\CoversClass;
+use PHPUnit\Framework\TestCase;
+
+#[CoversClass(PrePackagedSourceAssetName::class)]
+final class PrePackagedSourceAssetNameTest extends TestCase
+{
+    public function testPackageNames(): void
+    {
+        self::markTestIncomplete('todo'); // @todo
```
asgrim, line 16: Test to be completed

src/Platform/PrePackagedSourceAssetName.php
```diff
@@ -0,0 +1,42 @@
...
+                $package->extensionName()->name(),
+                $package->version(),
+            )),
+            strtolower(sprintf(
+                'php_%s-%s-src.zip',
+                $package->extensionName()->name(),
+                $package->version(),
+            )),
+            // @todo remove this:
+            strtolower(sprintf(
```
asgrim, line 35: To be removed; added only for testing with existing releases, but as discussed in #39 the existing releases could potentially be updated with the new `php_mongodb-x.y.z-src.zip` format.

resources/composer-json-php-ext-schema.json
```diff
@@ -0,0 +1,106 @@
...
+                    "example": false,
+                    "default": true
+                },
+                "build-path": {
+                    "type": ["string", "null"],
+                    "description": "If specified, this is the subdirectory that will be used to build the extension instead of the root of the project.",
+                    "example": "my-extension-source",
+                    "default": null
+                },
+                "override-download-url-method": {
```
asgrim, line 43: Corresponding PR made to add this to upstream Composer: composer/composer#12296

docs/extension-maintainers.md
```diff
@@ -160,6 +160,26 @@ should specify this path in `build-path`, for example:
...
 
+The `build-path` may contain some templated values which are replaced:
+
+ * `{version}` to be replaced with the package version. For example a package
+   with version 1.2.3 with a `build-path` of `myext-{version}` the actual build
+   path would become `myext-{version}`.
+
+#### `override-download-url-method`
+
+The `override-download-url-method` directive allows extension maintainers to
```
stof, line 171: should it have `override` in the name ?

asgrim: I figured it might be better, since you're overriding the default behaviour (i.e. use what Composer does). If you think it's superfluous, I'm happy to remove it? :thinking:

stof: Well, *any* configuration option overrides the default behavior when you specify a value. That's the reason to make it configurable.

asgrim: Upstream PR sent - thank you composer/composer#12297

src/DependencyResolver/Package.php
```diff
@@ -26,53 +27,64 @@
...
+        private readonly string|null $downloadUrl,
     ) {
+        $this->configureOptions       = [];
+        $this->supportZts             = true;
+        $this->supportNts             = true;
+        $this->buildPath              = null;
+        $this->compatibleOsFamilies   = null;
+        $this->incompatibleOsFamilies = null;
+        $this->priority               = 80;
+        $this->downloadUrlMethod      = null;
```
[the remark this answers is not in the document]
asgrim: Ah yeah they probably can be now; this was leftover from mucking around with `readonly` stuff, thanks!

composer.json
```diff
@@ -28,7 +28,7 @@
     ],
     "require": {
         "php": "8.1.*||8.2.*||8.3.*||8.4.*",
-        "composer/composer": "^2.8.5",
+        "composer/composer": "dev-main",
```
asgrim, line 31: We should probably pin to the next stable Composer release, when that happens :+1:

Seldaek: ^2.8.6@dev

docs/extension-maintainers.md
```diff
@@ -160,6 +160,26 @@ should specify this path in `build-path`, for example:
 }
 ```
 
+The `build-path` may contain some templated values which are replaced:
+
+ * `{version}` to be replaced with the package version. For example a package
+   with version 1.2.3 with a `build-path` of `myext-{version}` the actual build
+   path would become `myext-1.2.3`.
```
GromNaN, line 167: The `build-path` could be detected: if the root directory doesn't contain a `configure` file but there is a single directory that contains this file, then go into it.

For the [mongodb extension](https://github.com/mongodb/mongo-php-driver), the build-path is at the root of the Git repository, so this setting seems incompatible with installation from Git clone.

asgrim: Good idea! Leave it with me and I'll see what we can do, thank you

asgrim: Added, thanks for the suggestion! :heavy_check_mark:

composer.json
asgrim: Updated to use `^2.8.6` now it is released :)

docs/extension-maintainers.md
GromNaN: I've read the code, I don't understand how you did it. Could you explain it to me please?

asgrim: Of course! The "interjection" point here (and the path that the build + install processes use) is in `\Php\Pie\Downloading\DownloadedPackage`; specifically for this change, I made  `\Php\Pie\Downloading\DownloadedPackage::unfoldUnarchivedSourcePaths`; it will check for the existence of `config.m4` or `config.w32`, and if one was not found, it will try to change the source path (e.g. originally `vendor/mongodb/mongodb-extension` as that is where Composer downloads things) automatically to the asset/archive name (minus file extension), e.g. `vendor/mongodb/mongodb-extension/mongodb-1.20.0/config.[m4|w32]` and if it finds them, it will use that as the source path automatically. If it fails to find `config.m4`/`.w32` it will attempt to use the original path instead. So ultimately, the key is the existence of `config.m4` or `config.w32`.

The existing `build-path` override was just separated into the separate `\Php\Pie\Downloading\DownloadedPackage::overrideSourcePathUsingBuildPath` method for tidy-ness, and `build-path` is no longer used, at least for MongoDB (although I kept the `{version}` replacement in case anyone finds it useful). The `build-path` directive may be useful in general, e.g. if the PHP extension part is not in the "root" of the repository (for examlpe in the [protobuf](https://github.com/protocolbuffers/protobuf/tree/main/php/ext/google/protobuf) extension).

PR #199, opened 2025-03-06
Fix Windows reinstall/uninstall issue

### LINKED ISSUE #190: Reinstalling or Uninstalling Windows extensions fails
Reinstalling or uninstalling an extension in Windows does not work currently.

This is because when DLL is loaded in Windows, we cannot unload it from PHP in order to free Windows' lock on the file. If you attempt to overwrite or uninstall the DLL in Windows, you will get an unhelpful/cryptic error like:

> ```
> PHP Warning:  copy(C:\tools\php\ext\php_example_pie_extension.dll): Failed to
> open stream: Resource temporarily unavailable in src\Installing\WindowsInstall.php on line 127
> ```

(this specific error is trying to re-install an ext that is already installed)

After some searching, it seems this is because the DLL is loaded and therefore cannot be "written" whilst loaded. One potential solution here is to relaunch PHP without the extension loaded. I believe Composer has a mechanism to reload without `xdebug` (for performance) so it may be possible to leverage that.

asgrim:
Fixes #190

### DIFF
--- a/src/File/WindowsDelete.php
+++ b/src/File/WindowsDelete.php
@@ -1,0 +1,47 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Php\Pie\File;
+
+use Php\Pie\Util\Process;
+use RuntimeException;
+
+use function file_exists;
+use function sprintf;
+use function sys_get_temp_dir;
+use function tempnam;
+
+/** @internal This is not public API for PIE, so should not be depended upon unless you accept the risk of BC breaks */
+final class WindowsDelete
+{
+    /**
+     * Windows will hold a lock on the DLL, so we can't actually replace or
+     * delete the extension. However, Windows lets us move it out the way,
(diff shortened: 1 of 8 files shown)


src/File/WindowsDelete.php
```diff
@@ -0,0 +1,47 @@
...
+
+/** @internal This is not public API for PIE, so should not be depended upon unless you accept the risk of BC breaks */
+final class WindowsDelete
+{
+    /**
+     * Windows will hold a lock on the DLL, so we can't actually replace or
+     * delete the extension. However, Windows lets us move it out the way,
+     * so move it to the temp directory. It can then be cleaned up later.
+     */
+    public static function usingMoveToTemp(string $filename): void
```
asgrim, line 23: Note: not ideal, as it leaves a dangling temp file, but not sure a way around the lock on the file. Any suggestions to improve this are welcomed!

PR #201, opened 2025-03-11
composer-json-php-ext-schema.json: update `$id` to correct path

DanielEScherzer:
Ensure that the URI given leads to the actual file rather than a 404 by fixing the end from `main/composer-json-php-ext-schema.json` to `main/resources/composer-json-php-ext-schema.json` since the file is located in the `resources` directory.


asgrim (approved):

asgrim:
Thanks @DanielEScherzer :+1:

PR #202, opened 2025-03-12
Added PIE version banner to all commands

asgrim:
Adds a version banner to all commands, e.g.:

```
$ bin/pie install asgrim/example-pie-extension
🥧 PHP Installer for Extensions (PIE), dev-main, from The PHP Foundation
This command may need elevated privileges, and may prompt you for your password.
You are running PHP 8.3.17
Target PHP installation: 8.3.17 nts, on Linux/OSX/etc x86_64 (from /usr/bin/php8.3)
Found package: asgrim/example-pie-extension:2.0.2 which provides ext-example_pie_extension
```

### DIFF
--- a/bin/pie
+++ b/bin/pie
@@ -25,7 +25,10 @@ include $_composer_autoload_path ?? __DIR__ . '/../vendor/autoload.php';
 
 $container = Container::factory();
 
-$application = new Application('🥧 PHP Installer for Extensions (PIE)', PieVersion::get());
+$pieAppName = '🥧 PHP Installer for Extensions (PIE)';
+$pieVersion = PieVersion::get();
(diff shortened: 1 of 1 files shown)


stof:
shouldn't this be sent to STDERR instead, to avoid breaking the usage of piping the command output to another process ?

remicollet:
> shouldn't this be sent to STDERR instead, to avoid breaking the usage of piping the command output to another process ?

Yes it will work (for bash completion)
Another way is to display the banner only if a tty is used (no redirect)

asgrim:
At the moment we are not really promising a stable textual output "API", but the feedback is noted. #217 will improve this anyway 👍

PR #205, opened 2025-03-24
Add self-update functionality

### LINKED ISSUE #66: PIE self-update command
Similar to `composer self-update`, we should add a `pie self-update` command.

asgrim:
Fixes #66


src/SelfManage/Verify/FallbackVerificationUsingOpenSsl.php
```diff
@@ -0,0 +1,302 @@
...
+                    }
+
+                    // Verify the CA cert has not expired
+                    Assert::keyExists($caCertificateInfo, 'validFrom_time_t');
+                    Assert::integer($caCertificateInfo['validFrom_time_t']);
+                    Assert::keyExists($caCertificateInfo, 'validTo_time_t');
+                    Assert::integer($caCertificateInfo['validTo_time_t']);
+
+                    $caValidFrom = new DateTimeImmutable('@' . $caCertificateInfo['validFrom_time_t']);
+                    $caValidTo   = new DateTimeImmutable('@' . $caCertificateInfo['validTo_time_t']);
```
TimWolla, line 190: I'm not sure if we should validate this: For code signing, the CA is generally only required to be valid at the time of *signing*, because the resulting binary might be used well in excess of the CA’s lifetime.

To ensure that a signing certificate is not used after its expiry to sign new binaries, a trusted timestamp is applied to the resulting binary. In case of the GH attestation, this would be the Sigstore logs.

I would find it reasonable to not validate the `caValidTo` and neither check against the Sigstore logs, because we don't or can't easily check against an revoked CA certificate either. However we must ensure that for every new PIE version, we clean out any expired CA certificates from the trust store.

Alternatively: Validate `caValidTo` not against `$this->now`, but against the build time of the PIE binary we are updating from and `caValidFrom` both against the build time and the current time.

So if we release PIE 0.9 on 2025-04-01 and PIE 0.10 on 2025-05-01 and the user updates on 2025-07-01 from 0.9 to 0.10, then we would consider all CA certificates that are valid on April, 1st as valid and for the next update all certificates that are valid on May, 1st.

TimWolla: Given that you download the attestation from GitHub anyways, you can also validate the `.tlogEntries.integratedTime`, which is the time that Sigstore claims the artifact was signed. So we would have the following constraints:

- CA_from < build_time(current_pie) < NOW(): The CA must be created before the current PIE binary and the current PIE binary must be created before the current time.
- build_time(current_pie) < CA_to: The CA must not be already expired when the PIE binary was created.
- CA_from < Attestation_from: The CA must be created before the signing certificate.
- CA_from < .tlogEntries.integratedTime: The CA must be created before the trusted time of signing.
- Attestation_to < CA_to: The signing certificate must not outlive the CA.
- .tlogEntries.integratedTime < CA_to: The CA must be unexpired at the trusted time of signing.
- .tlogEntries.integratedTime should “roughly match” match the validity period of the Attestation certificate. I'm not sure what guarantees are provided by Sigstore, but for the public transparency logs for HTTPS there is a “maximum merge delay” of 24 hours, which means that all entries in the log must be up to date within 24 hours. So I'd say that .tlogEntries.integratedTime should be no more that 24 hours before Attestation_from and no more than 24 hours after Attestation_to.

asgrim: Perhaps lets just remove the CA valid time then; I'm keen to draw a line "somewhere" as long as we're doing a basic check that the PHAR is sourced from the correct place.

src/SelfManage/Verify/FallbackVerificationUsingOpenSsl.php
```diff
@@ -0,0 +1,302 @@
...
+                        || ! is_string($caCertificateWrapper['rawBytes'])
+                        || $caCertificateWrapper['rawBytes'] === ''
+                    ) {
+                        continue;
+                    }
+
+                    // Format the certificate, since OpenSSL doesn't accept the rawBytes directly
+                    $caCertificateString = "-----BEGIN CERTIFICATE-----\n"
+                        . wordwrap($caCertificateWrapper['rawBytes'], 67, "\n", true) . "\n"
+                        . "-----END CERTIFICATE-----\n";
```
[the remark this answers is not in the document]
asgrim: Updated

src/SelfManage/Verify/FallbackVerificationUsingOpenSsl.php
```diff
@@ -0,0 +1,302 @@
...
+            Assert::stringNotEmpty($attestationCertificateInfo['extensions'][$extension]);
+            $actualValue = $attestationCertificateInfo['extensions'][$extension];
+
+            // First character (the ASN.1 tag) is expected to be UTF8String (0x0C)
+            if (ord($actualValue[0]) !== 12) {
+                throw FailedToVerifyRelease::fromMismatchingExtensionValues($extension, $expectedValue, $actualValue);
+            }
+
+            // Second character is expected to be the length of the actual value
+            $derDecodedValue = mb_substr($actualValue, 2, ord($actualValue[1]), '8bit');
```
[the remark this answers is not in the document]
asgrim: Changed to use `substr`

src/SelfManage/Verify/FallbackVerificationUsingOpenSsl.php
TimWolla: > Perhaps lets just remove the CA valid time then

Yes, I agree that this is a reasonable compromise, as long as the list of certificates is pruned from expired certificates at build time (e.g. by aGitHub Action that updates the certificate file regularly and sends PRs).

asgrim: Created #212

src/SelfManage/Verify/FallbackVerificationUsingOpenSsl.php
```diff
@@ -0,0 +1,297 @@
...
+                throw FailedToVerifyRelease::fromMismatchingExtensionValues($extension, $expectedValue, $actualValue);
+            }
+
+            /**
+             * Second character is expected to be the length of the actual value
+             * as long as they are less than 127 bytes (short form)
+             *
+             * @link https://www.oss.com/asn1/resources/asn1-made-simple/asn1-quick-reference/basic-encoding-rules.html#Lengths
+             */
+            $derDecodedValue = substr($actualValue, 2, ord($actualValue[1]));
```
TimWolla, line 120: Please add a check that the provided length actually matches the length of the contents. This is simple to do and prevents some potential shenanigans with the claimed length to be shorter than the length that OpenSSL extracts.

So basically:

```
"\x0c\x0512345678"
```

This claims to a length of 5, but the actual payload is 8 bytes in length. That should be rejected. And likewise should:

```
"\x0c\x05123"
```

I gave an example in https://github.com/php/pie/pull/205#discussion_r2022905578, but I guess you missed it, because I didn't mention it in the explanation (just in the suggested diff).

src/SelfManage/Verify/FallbackVerificationUsingOpenSsl.php
```diff
@@ -0,0 +1,297 @@
...
+        Assert::isArray($attestationCertificateInfo['extensions']);
+
+        /**
+         * See {@link https://github.com/sigstore/fulcio/blob/main/docs/oid-info.md#136141572641--fulcio} for details
+         * on the Fulcio extension keys; note the values are DER-encoded strings; the ASN.1 tag is UTF8String (0x0C).
+         *
+         * First up, check the extension values are what we expect; these are hard-coded, as we don't expect them
+         * to change unless the namespace/repo name change, etc.
+         */
+        foreach (self::ATTESTATION_CERTIFICATE_EXPECTED_EXTENSION_VALUES as $extension => $expectedValue) {
```
TimWolla, line 104: This should move to the bottom: It's a more logical verification order.

1. Verify that the certificate is a validly signed certificate in the first place.
2. Then verify that the certificate belongs to PIE.

Trying to extract data from an unverified certificate is questionable and can lead to confusing error messages).

TimWolla (commented):
LGTM, except for the two small remarks.

src/SelfManage/Verify/FallbackVerificationUsingOpenSsl.php
```diff
@@ -0,0 +1,297 @@
...
+         * And verifying the contents afterwards to ensure they have not been compromised. This list of JSON blobs may
+         * have multiple certificates (e.g. root certificates, intermediate certificates, expired certificates, etc.)
+         * so we should loop over to find the correct certificate used to sign the attestation certificate.
+         */
+        foreach ($trustedRootJsonLines as $jsonLine) {
+            /** @var mixed $decoded */
+            $decoded = json_decode($jsonLine, true);
+
+            // No certificate authorities defined in this JSON line, skip it...
+            if (! is_array($decoded) || ! array_key_exists('certificateAuthorities', $decoded)) {
```
TimWolla, line 144:
```suggestion
            if (! is_array($decoded) || ! array_key_exists('certificateAuthorities', $decoded) || ! is_array($decoded['certificateAuthorities'])) {
```
?

src/SelfManage/Verify/FallbackVerificationUsingOpenSsl.php
asgrim: Makes sense, updated

src/SelfManage/Verify/FallbackVerificationUsingOpenSsl.php
asgrim: Ah yes, I did miss that, thank you! :+1:

TimWolla (approved):

PR #213, opened 2025-04-09
Add "pie install" to install missing extensions for a PHP project

asgrim:
Add `pie install` without parameters to check a PHP project for missing extensions

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,6 +49,36 @@ You must now add "extension=example_pie_extension" to your php.ini
 $
 ```
 
+## Installing all extensions for a project
+
+When in your PHP project, you can install any missing top-level extensions:
+
+```
+$ pie install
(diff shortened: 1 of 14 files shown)


stof:
It would be great to support identifying all extensions required by a project when a composer.lock file is present.

the heroku PHP buildpack has such logic looking at the lock file: https://github.com/heroku/heroku-buildpack-php/blob/6af5a55aab8a8f9417b8c8a14b0a6f91bf005aa0/bin/util/platform.php#L124

asgrim:
Yep thanks for the suggestion, that is already in the works 👍

asgrim:
> It would be great to support identifying all extensions required by a project when a composer.lock file is present.

ticket created for tracking #222

PR #223, opened 2025-04-17
Self verify command

asgrim:
Using the existing verification code, we can run `pie self-verify` to check we're running an authentic PIE

### DIFF
--- a/src/Command/SelfVerifyCommand.php
+++ b/src/Command/SelfVerifyCommand.php
@@ -1,0 +1,89 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Php\Pie\Command;
+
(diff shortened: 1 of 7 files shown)


src/Command/SelfVerifyCommand.php
```diff
@@ -0,0 +1,89 @@
...
+    {
+        parent::configure();
+
+        CommandHelper::configurePhpConfigOptions($this);
+    }
+
+    public function execute(InputInterface $input, OutputInterface $output): int
+    {
+        if (! PieVersion::isPharBuild()) {
+            $output->writeln('<comment>Aborting! You are not running a PHAR, cannot self-update.</comment>');
```
TimWolla, line 51: s/update/verify/?

PR #235, opened 2025-05-05
Create a GH issue when trusted root certs are outdated

### LINKED ISSUE #212: Automate maintenance of trusted roots
In #205 we added a `trusted-roots.jsonl` which is updated by running:

```bash
gh attestation trusted-root > resources/trusted-root.jsonl
```

We could automate;

 * making a PR to update the file
 * making a PR to remove expired certificates

asgrim:
Fixes #212

### DIFF
--- a/.github/workflows/check-outdated-trusted-root.yml
+++ b/.github/workflows/check-outdated-trusted-root.yml
@@ -2,7 +2,7 @@ name: "Check for outdated trusted root certificates"
 
 on:
   schedule:
-    - cron: "0 0 * * 2"
+    - cron: "0 0 * * *"
 
 jobs:
   check:
--- a/.github/workflows/check-outdated-trusted-root.yml
+++ b/.github/workflows/check-outdated-trusted-root.yml
@@ -17,4 +17,8 @@ jobs:
         run: |
           diff resources/trusted-root.jsonl resources/new-trusted-root.jsonl \
             && echo "Trusted root cert has not changed, no action required." \
-            || gh issue create -R php/pie --title "Trusted root needs updating" --body "The trusted root certificate file needs updating. Use the \`gh attestation trusted-root > resources/trusted-root.jsonl\` command to update it." --assignee "asgrim"
+            || ( \
+              (gh issue list | grep -i "Trusted root needs updating") \
+              && echo "Issue to update trusted root already exists, no action required." \
+              || gh issue create --title "Trusted root needs updating" --body "The trusted root certificate file needs updating. Use the \`gh attestation trusted-root > resources/trusted-root.jsonl\` command to update it." --assignee "asgrim" \
+            )
--- a/.github/workflows/check-outdated-trusted-root.yml
+++ b/.github/workflows/check-outdated-trusted-root.yml
@@ -1,0 +1,20 @@
+name: "Check for outdated trusted root certificates"
+
+on:
+  schedule:
+    - cron: "0 0 * * 2"
+
+jobs:
+  check:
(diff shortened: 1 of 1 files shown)


.github/workflows/check-outdated-trusted-root.yml
```diff
@@ -0,0 +1,20 @@
...
+      - uses: actions/checkout@v4
+      - name: Pull new trusted root
+        run: gh attestation trusted-root > resources/new-trusted-root.jsonl
+      - name: Create an issue if different
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          diff resources/trusted-root.jsonl resources/new-trusted-root.jsonl \
+            && echo "Trusted root cert has not changed, no action required." \
+            || gh issue create -R php/pie --title "Trusted root needs updating" --body "The trusted root certificate file needs updating. Use the \`gh attestation trusted-root > resources/trusted-root.jsonl\` command to update it." --assignee "asgrim"
```
asgrim, line 20: Need to add a check if there is already an open issue with same name to avoid re-opening

.github/workflows/check-outdated-trusted-root.yml
```diff
@@ -0,0 +1,20 @@
+name: "Check for outdated trusted root certificates"
+
+on:
+  schedule:
+    - cron: "0 0 * * 2"
```
asgrim, line 5:
```suggestion
    - cron: "0 0 * * *"
```

.github/workflows/check-outdated-trusted-root.yml
```diff
@@ -0,0 +1,24 @@
+name: "Check for outdated trusted root certificates"
+
+on:
+  schedule:
+    - cron: "0 0 * * *"
+
+jobs:
```
TimWolla, line 7: Missing permissions section: I believe all repositories in the php/* organization are configured with read-only GHA permissions by default. Thus this is unable to create the issue. You'll need:

```
permissions:
  issues: write
```

see: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/controlling-permissions-for-github_token

.github/workflows/check-outdated-trusted-root.yml
```diff
@@ -0,0 +1,24 @@
...
+      - name: Pull new trusted root
+        run: gh attestation trusted-root > resources/new-trusted-root.jsonl
+      - name: Create an issue if different
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          diff resources/trusted-root.jsonl resources/new-trusted-root.jsonl \
+            && echo "Trusted root cert has not changed, no action required." \
+            || ( \
+              (gh issue list | grep -i "Trusted root needs updating") \
```
TimWolla, line 21: I think this might not work if too many open issues exist. This should not happen in practice, but perhaps you could make this a little smarter by leveraging the filter flags: https://cli.github.com/manual/gh_issue_list

Something: `gh issue list --author "github-actions[bot]"` might help. Or just create a dedicated label.

.github/workflows/check-outdated-trusted-root.yml
asgrim: Thanks! Good catch, tested it on my own fork, but that doesn't have that restriction I think :) #237

TimWolla: Quite possible. Organization-owned repositories behave a little differently. You can sync up the permissions here:

image

That would likely be at https://github.com/asgrim/pie/settings/actions

PR #237, opened 2025-05-06
Add permissions for creating issues to trusted root cert check job

asgrim:
Fix for https://github.com/php/pie/pull/235#discussion_r2076288514

### DIFF
--- a/.github/workflows/check-outdated-trusted-root.yml
+++ b/.github/workflows/check-outdated-trusted-root.yml
@@ -21,7 +21,7 @@ jobs:
           diff resources/trusted-root.jsonl resources/new-trusted-root.jsonl \
             && echo "Trusted root cert has not changed, no action required." \
             || ( \
-              (gh issue list | grep -i "Trusted root needs updating") \
+              (gh issue list --label "trusted-root-update" | grep -i "Trusted root needs updating") \
               && echo "Issue to update trusted root already exists, no action required." \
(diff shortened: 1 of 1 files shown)


.github/workflows/check-outdated-trusted-root.yml
```diff
@@ -7,6 +7,9 @@ on:
 jobs:
   check:
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      issues: write
```
TimWolla, line 12: Please also add a top-level `permissions` section. For consistency with the other workflows and also because it is a little "safer" if additional jobs are added after-the-fact. The top-level `permissions` could just be `contents: read`, the more specific job-level permissions will override this.

TimWolla: See #127 for further context (that's where I initially introduced the permissions).

TimWolla, line 12:
```suggestion
      # issues:write is required to create issues.
      issues: write
```
The other `write` permissions have a short explanatory comment. Here it's a little tautological, but for consistency 😄

TimWolla (commented):
Already merged, but since you requested a review, LGTM.

asgrim:
> Already merged, but since you requested a review, LGTM.

Thanks :) wanted to get it merged before scheduled job just to make sure it was all happy (it was!) :)

PR #262, opened 2025-06-18
Documentation improvements

asgrim:
From a couple of bits of feedback after 1.0 release.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ On Linux, you will need a build toolchain installed. On Debian/Ubuntu type
 systems, you could run something like:
 
 ```shell
-sudo apt install build-essential autoconf libtool bison re2c pkg-config
+sudo apt install gcc make autoconf libtool bison re2c pkg-config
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,8 +1,31 @@
...
+## What do I need to get started?
+
 You will need PHP 8.1 or newer to run PIE, but PIE can install an extension to
 any installed PHP version.
 
+On Linux, you will need a build toolchain installed. On Debian/Ubuntu type
+systems, you could run something like:
+
+```shell
+sudo apt install build-essential autoconf libtool bison re2c pkg-config
```
TimWolla, line 20: build-essential is a meta package to build Debian packages. It is not correct to list here (it's a common mistake, due to the bad name of the package). 
```suggestion
sudo apt install gcc make autoconf libtool bison re2c pkg-config
```

This should (hopefully) be sufficient to have everything needed to build extensions.

PR #275, opened 2025-07-03
133: Add bundled php exts support

### LINKED ISSUE #133: Make core bundled extensions available (PDO, curl, intl, etc.)
Ref: https://bsky.app/profile/nizzness.bsky.social/post/3lbntufzx6c2i

> Anyone know if the pie installer has plans to add things like intl, pdo, curl etc...?

Extensions in the core such as PDO, curl, intl currently aren't available in PIE. Discuss how to make them available, since a root composer.json won't really work, as it's essentially one repo with a whole bunch of extensions.

asgrim:
Fixes #133

### DIFF
--- a/test/unit/ComposerIntegration/BundledPhpExtensionsRepositoryTest.php
+++ b/test/unit/ComposerIntegration/BundledPhpExtensionsRepositoryTest.php
@@ -1,0 +1,203 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Php\PieUnitTest\ComposerIntegration;
+
+use Composer\Package\CompletePackageInterface;
+use Php\Pie\ComposerIntegration\BundledPhpExtensionsRepository;
+use Php\Pie\DependencyResolver\Package;
+use Php\Pie\Downloading\DownloadedPackage;
+use Php\Pie\ExtensionName;
+use Php\Pie\ExtensionType;
+use Php\Pie\Platform\Architecture;
+use Php\Pie\Platform\OperatingSystem;
+use Php\Pie\Platform\OperatingSystemFamily;
+use Php\Pie\Platform\TargetPhp\PhpBinaryPath;
+use Php\Pie\Platform\TargetPlatform;
+use Php\Pie\Platform\ThreadSafetyMode;
+use Php\Pie\Util\Process;
+use PHPUnit\Framework\Attributes\CoversClass;
+use PHPUnit\Framework\Attributes\DataProvider;
+use PHPUnit\Framework\TestCase;
+use Symfony\Component\Process\Exception\ProcessFailedException;
+
+use function mkdir;
+use function realpath;
+use function sys_get_temp_dir;
+use function uniqid;
+
+use const DIRECTORY_SEPARATOR;
+
+#[CoversClass(BundledPhpExtensionsRepository::class)]
+final class BundledPhpExtensionsRepositoryTest extends TestCase
+{
+    /** @return list<array{0: non-empty-string}> */
+    public static function bundledRepositoryPackageNames(): array
+    {
+        return [
+            ['php/bcmath'],
+            ['php/bz2'],
+            ['php/calendar'],
+            ['php/ctype'],
+            ['php/curl'],
+            ['php/dba'],
+            ['php/dom'],
+            ['php/enchant'],
+            ['php/exif'],
+            ['php/ffi'],
+            ['php/gettext'],
+            ['php/gmp'],
+            ['php/iconv'],
+            ['php/intl'],
+            ['php/ldap'],
+            ['php/mbstring'],
+            ['php/mysqlnd'],
+            ['php/mysqli'],
+            ['php/opcache'],
+            ['php/pcntl'],
+            ['php/pdo'],
+            ['php/pdo_mysql'],
+            ['php/pdo_pgsql'],
+            ['php/pdo_sqlite'],
+            ['php/pgsql'],
+            ['php/posix'],
+            ['php/readline'],
+            ['php/session'],
+            ['php/shmop'],
+            ['php/simplexml'],
+            ['php/snmp'],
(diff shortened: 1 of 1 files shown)


src/ComposerIntegration/BundledPhpExtensionsRepository.php
```diff
@@ -0,0 +1,150 @@
...
+        ['name' => 'soap'],
+        ['name' => 'sockets'],
+        ['name' => 'sodium', 'version' => '>= 7.2.0'],
+        ['name' => 'sqlite3', 'version' => '>= 5.3.0'],
+        ['name' => 'sysvmsg'],
+        ['name' => 'sysvsem'],
+        ['name' => 'sysvshm'],
+        ['name' => 'tidy'],
+        // ['name' => 'tokenizer'], // build failure - make: *** No rule to make target '/home/james/workspace/oss/php-src/ext/tokenizer/Zend/zend_language_parser.y', needed by '/home/james/workspace/oss/php-src/ext/tokenizer/Zend/zend_language_parser.c'. Stop.
+        ['name' => 'uri', 'version' => '>= 8.5.0'],
```
TimWolla, line 92: This extension is always loaded.

src/ComposerIntegration/BundledPhpExtensionsRepository.php
```diff
@@ -0,0 +1,150 @@
...
+            'name' => 'mysqli',
+            'priority' => 90, // must load after mysqlnd
+        ],
+        ['name' => 'mysqlnd', 'version' => '>= 5.3.0'],
+        // ['name' => 'odbc'], // build failure - cp: cannot stat '/usr/local/lib/odbclib.a': No such file or directory configure: error: ODBC header file '/usr/local/incl/sqlext.h' not found!
+        [
+            'name' => 'opcache',
+            'type' => ExtensionType::ZendExtension,
+            'version' => '>= 5.5.0',
+        ],
```
TimWolla, line 65: This extension is always loaded.

src/ComposerIntegration/BundledPhpExtensionsRepository.php
```diff
@@ -0,0 +1,150 @@
...
+        ['name' => 'bz2'],
+        ['name' => 'calendar'],
+        [
+            'name' => 'com_dotnet',
+            'os-families' => [OperatingSystemFamily::Windows],
+        ],
+        ['name' => 'ctype'],
+        ['name' => 'curl'],
+        ['name' => 'dba'],
+        ['name' => 'dl_test', 'version' => '>= 8.2.0'],
```
TimWolla, line 42: This is not useful to expose to the user.

src/ComposerIntegration/BundledPhpExtensionsRepository.php
```diff
@@ -0,0 +1,150 @@
...
+        ['name' => 'sysvsem'],
+        ['name' => 'sysvshm'],
+        ['name' => 'tidy'],
+        // ['name' => 'tokenizer'], // build failure - make: *** No rule to make target '/home/james/workspace/oss/php-src/ext/tokenizer/Zend/zend_language_parser.y', needed by '/home/james/workspace/oss/php-src/ext/tokenizer/Zend/zend_language_parser.c'. Stop.
+        ['name' => 'uri', 'version' => '>= 8.5.0'],
+        ['name' => 'xml'],
+        ['name' => 'xmlreader', 'version' => '>= 5.1.0'],
+        ['name' => 'xmlwriter', 'version' => '>= 5.2.0'],
+        ['name' => 'xsl'],
+        // ['name' => 'zend_test', 'version' => '>= 7.2.0'], // build failure - ext/zend_test/test.c:48:11: fatal error: libxml/globals.h: No such file or directory
```
TimWolla, line 97: This is not useful to expose to the user.

src/ComposerIntegration/BundledPhpExtensionsRepository.php
```diff
@@ -0,0 +1,150 @@
...
+        ['name' => 'enchant', 'version' => '>= 5.2.0'],
+        ['name' => 'exif'],
+        ['name' => 'ffi', 'version' => '>= 7.4.0'],
+        ['name' => 'gd'],
+        ['name' => 'gettext'],
+        ['name' => 'gmp'],
+        ['name' => 'iconv'],
+        ['name' => 'intl', 'version' => '>= 5.3.0'],
+        ['name' => 'ldap'],
+        ['name' => 'lexbor', 'version' => '>= 8.5.0'],
```
TimWolla, line 53: This is always loaded with PHP 8.5.

TimWolla (changes requested):
Besides the remarks, I'm more generally concerned about this. As you can see for the extensions where you commented the build failures, the bundled extensions are not tested outside of PHP’s build system and it's probably easy to accidentally choose some different `configure` options that result in subtly broken extensions, especially when considering that Linux distributions often ship with patched PHP versions, where PIE is going to miss out on those patches. The bundled extensions really should just be installed by the respective package manager.

asgrim:
@TimWolla

> As you can see for the extensions where you commented the build failures, the bundled extensions are not tested outside of PHP’s build system

Yes, my plan here is to add an integration test that does a test `pie install <package>` for each of these defined ones. I'm working on that next so you'll see what I mean shortly

> The bundled extensions really should just be installed by the respective package manager.

This was my expectation too, but we have had several requests, I think the most common would be PDO*, bcmath, mbstring.

src/ComposerIntegration/BundledPhpExtensionsRepository.php
TimWolla: Sorry, this meant to read "is always loaded with PHP 8.5". This is *not yet* the case, but will likely be with 8.5.0 Alpha 3 as part of https://wiki.php.net/rfc/make_opcache_required.

PR #277, opened 2025-07-08
Added PHP 8.5 support

thecaliskan:
This pull request adds support for PHP 8.5 to the package.

- Adjusted compatibility to allow installation on PHP 8.5
- Verified that existing functionality works as expected under PHP 8.5

Please let me know if any changes or additional testing are required.

Thank you!

PR #281, opened 2025-07-17
Regenerate composer.lock

TimWolla:
php/pie#277 adjusted composer.json without also updating composer.json.


thecaliskan:
I forgot to run the update command. Thank you for contributing.

PR #282, opened 2025-07-17
Run `composer validate --strict` as part of the Static Analysis CI

TimWolla:
CI is expected to be red until #281 is merged.

----------------

This will ensure that the `composer.json` is in sync with `composer.lock` and thus is related to php/pie#281 and php/pie#277.


asgrim:
Thanks @TimWolla !

PR #328, opened 2025-08-27
docs: add php-dev as dependency on Debian/Ubuntu systems

chapeupreto:
This commit adds `php-dev` as a dependency on Debian/Ubuntu systems, so the following error message is fixed:

image


asgrim (approved):
Thanks @chapeupreto

PR #349, opened 2025-09-03
Add nightly release

alexandre-daubois:
Fix #323 

Essentially the same as `release.yml`, but scheduled to run every day at 2 am UTC.

PR #350, opened 2025-09-03
De-duplicate docker image build for nightly and release tags

### LINKED ISSUE #323: Can we push `pie:nightly` docker image
push nightly version pie docker image，we can use latest fixed version.

### LINKED ISSUE #334: RC versions are published as  `ghcr.io/php/pie:bin`
RC versions should probably not be published to the `bin` alias of the Docker Image, because otherwise folks will accidentally consume it. They should only be available by an explicit version request.

I'm also noticing that the Docker image tag for RC versions does not include the `-bin` suffix. 1.2.0 RC 1 is available as `1.2.0-rc.1` instead of `1.2.0-rc.1-bin` or similar: https://github.com/php/pie/pkgs/container/pie/498699978?tag=1.2.0-rc.1

asgrim:
Fixes #323

Fixes #334 

Supersedes #349 

Sorry for splitting into a new PR @alexandre-daubois - was quicker to just do this rather than provide direction as I already had a plan for this ticket. This approach just avoids all the duplication in #349  - also, it makes no sense IMO to actually run it every night on schedule, just whenever `main` is merged to (same as the nightly PHAR build) -

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -39,76 +39,20 @@ jobs:
         with:
           files: pie.phar
 
-  docker-binary-only-image:
-    needs: build-phar
-    name: Docker binary-only image
-    runs-on: ubuntu-latest
+  build-and-push-docker-image:
     if: ${{ startsWith(github.ref, 'refs/tags/') }}
-
+    needs: build-phar
+    # See build-and-push-docker-image.yml for a list of the permissions and why they are needed
     permissions:
-      # attestations:write is required for build provenance attestation.
-      attestations: write
-      # id-token:write is required for build provenance attestation.
+      contents: read
       id-token: write
-      # packages:write is required to publish Docker images to GitHub's registry.
+      attestations: write
       packages: write
-
-    steps:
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -39,76 +39,20 @@ jobs:
...
-          subject-name: ghcr.io/${{ github.repository }}
-          subject-digest: ${{ steps.build-and-push.outputs.digest }}
-          push-to-registry: true
+    uses: ./.github/workflows/build-and-push-docker-image.yml
+    with:
+      # @TODO v1.0 Consider introducing more granular tags (major and major.minor)
+      # @see https://github.com/php/pie/pull/122#pullrequestreview-2477496308
+      # @see https://github.com/php/pie/pull/122#discussion_r1867331273
+      tags: |
+        type=raw,value=bin
```
asgrim, line 57: Todo: don't publish `bin` tag when non-stable versions published

alexandre-daubois:
That's fine, thanks for taking care of it!

PR #402, opened 2025-10-31
Skip test with no known /usr/bin/php-configX.Y

mbeccati:
I was still getting random failures due to `php -i` taking more than 5 seconds and I tracked down the issue to a test using other available versions of php-config on the system.

Once removed the system php, this data provider returned an empty array, and phpunit is not happy about it.

Cheers

### DIFF
--- a/test/unit/Platform/TargetPhp/PhpBinaryPathTest.php
+++ b/test/unit/Platform/TargetPhp/PhpBinaryPathTest.php
@@ -135,6 +135,10 @@ final class PhpBinaryPathTest extends TestCase
                 && is_executable($phpConfigPath[0]),
         );
 
+        if ($possiblePhpConfigPaths === []) {
+            return ['skip' => ['skip', 'skip']];
+        }
+
         return array_combine(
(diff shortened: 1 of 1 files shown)


test/unit/Platform/TargetPhp/PhpBinaryPathTest.php
```diff
@@ -135,6 +135,10 @@ public static function phpConfigPathProvider(): array
                 && is_executable($phpConfigPath[0]),
         );
 
+        if ($possiblePhpConfigPaths === []) {
+            return ['skip' => ['skip', 'skip']];
```
asgrim: Wouldn't this still try to run a test; you would need to add a check for `skip` into `testFromPhpConfigExecutable`; the windows skip above only works because the `testFromPhpConfigExecutable` test has the exact same check. That check could maybe be changed to see `if ($phpConfigPath === 'skip') { self::markTestSkipped...` perhaps?

asgrim (approved):

PR #427, opened 2025-11-25
Add `relay` extension to supported extensions list

tillkruss:
See https://packagist.org/packages/cachewerk/ext-relay


asgrim (approved):
Thanks @tillkruss 🎉

PR #443, opened 2025-12-01
Experimentally provide executable version of PIE

### LINKED ISSUE #432: Executable static build of PIE+PHP
It may be possible to distribute PIE as a complete standalone binary, using [Static PHP](https://static-php.dev/) - i.e. PHP itself, the extensions that PIE depends on, and PIE itself.

The huge benefit for this is it eliminates the need for specific PHP versions on the system; currently PIE requires PHP 8.1, but that means to install an extension on an older PHP version, you must have two PHP versions (one to run PIE, plus your target) and use the `--with-php-config=...` option. If we build it all together, it is standalone and much more portable.

This has an added side benefit in the future of being able to run without ANY dependencies, and means we could "drop" older PHP versions for the PIE tool itself, eventually (assuming we drop the PHAR distribution).

asgrim:
Fixes #432


.github/workflows/build-assets.yml
```diff
@@ -0,0 +1,161 @@
...
+     - build-phar
+    runs-on: ${{ matrix.operating-system }}
+    strategy:
+      fail-fast: false
+      matrix:
+        operating-system:
+          - ubuntu-24.04
+          - ubuntu-24.04-arm
+          - macos-15-intel
+          - macos-26
```
asgrim, line 78: OSX binaries may be problematic to provide; a test someone with an ARM64 Mac revealed:

Image

I assume we'll have to look at signing it somehow. Having never used a Mac in anger, I'll need to look into how we can do this in CI (if possible).

We could potentially leave it as is, at least for now; I expect if/when we get more adoption of the executable distribution, we can revisit this separately.

.github/workflows/build-assets.yml
```diff
@@ -0,0 +1,161 @@
...
+          - windows-2025
```
asgrim, line 79: Weirdly, I couldn't get the `.exe` to run on my Windows VM. However, a couple of other folks have tried this on their Windows machines and have had good success, so I'm putting the issue down to some weird virtualisation thing, I guess.

src/Command/SelfUpdateCommand.php
```diff
@@ -82,7 +82,7 @@ public function configure(): void
 
     public function execute(InputInterface $input, OutputInterface $output): int
     {
-        if (! PieVersion::isPharBuild()) {
+        if (! PieVersion::isPharBuild() || Platform::isRunningStaticPhp()) {
```
asgrim, line 85: As this is initially an experiment to see if it is useful at all; we can split out adding self-update support separately.

.github/workflows/build-assets.yml
asgrim: I was given this to review - https://users.rust-lang.org/t/distributing-cli-apps-on-macos/70223/12 - may contain useful info

asgrim: Found this, may have useful references: https://github.com/simonw/datasette/issues/1171

shyim: Yes you will need an Developer Account which costs 100$ a year at Apple. I setuped it 3 months ago for a Go project, is a pain. See https://github.com/anchore/quill

If you need help let me now, but you would need first an Paid Apple Developer Account 😅

asgrim: Thanks; I think I may split this out and handle it separately for now. This is really just an experiment, but if people find it is useful to have the binaries, then the signing stuff is something we can add on later :)

asgrim: Created an "idea" thread here to handle later: php/pie#459

For now, Apple folks will have to trust it however that works.

src/Command/SelfUpdateCommand.php
asgrim: Moved to: php/pie#460

PR #450, opened 2025-12-08
Add docs translations for Chinese and Japanese

VinchanGit:


VinchanGit:
Hi asgrim, I see the task has been assigned to me. Is there any adjustment I need to make? @asgrim

asgrim:
> Hi asgrim, I see the task has been assigned to me. Is there any adjustment I need to make? @asgrim 

Nope, I need to figure this. My concern is keeping the translation up to date when other changes are made, and whether PIE's more permanent home will be on the php.net docs rather than here... So far it's not really been discussed. Not an issue with the change itself!

VinchanGit:
> > Hi asgrim, I see the task has been assigned to me. Is there any adjustment I need to make? @asgrim
> 
> Nope, I need to figure this. My concern is keeping the translation up to date when other changes are made, and whether PIE's more permanent home will be on the php.net docs rather than here... So far it's not really been discussed. Not an issue with the change itself!

i think , is not to merge it into php.net, because older versions of PHP do not support PIE. Additionally, it would make read the documentation less convenient. Moreover, merging it into php.net would feel somewhat bloated.

asgrim (approved):

PR #452, opened 2025-12-15
Prompt to install build tools if they're missing

### LINKED ISSUE #434: Prompt to install build tools (gcc/phpize/etc) if missing
Add a pre-check to ensure relevant toolchain is available for building.

By default, PIE should prompt to see if the user wants to install any missing tools. Example for Ubunutu `apt install gcc make autoconf libtool bison re2c pkg-config`; check for package names in yum/apk/etc too depending on the system. If we can't detect how to install the tools (or we are running non-interactively), just write a warning to the user that the tools are missing, and they should be installed.

However, this checking behaviour should be disabled with a flag, if provided, as some people may not want it.

asgrim:
Fixes #434

### DIFF
--- a/src/Command/CommandHelper.php
+++ b/src/Command/CommandHelper.php
@@ -251,6 +251,10 @@ final class CommandHelper
 
     public static function shouldCheckForBuildTools(InputInterface $input): bool
     {
+        if (Platform::isWindows()) {
+            return false;
+        }
+
         return ! $input->hasOption(self::OPTION_SUPPRESS_BUILD_TOOLS_CHECK)
             || ! $input->getOption(self::OPTION_SUPPRESS_BUILD_TOOLS_CHECK);
     }
--- a/src/Command/CommandHelper.php
+++ b/src/Command/CommandHelper.php
@@ -61,6 +61,8 @@ final class CommandHelper
     private const OPTION_SKIP_ENABLE_EXTENSION                = 'skip-enable-extension';
     private const OPTION_FORCE                                = 'force';
     private const OPTION_NO_CACHE                             = 'no-cache';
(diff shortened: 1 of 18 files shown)


src/Command/CommandHelper.php
```diff
@@ -228,6 +243,18 @@ public static function determineForceInstallingPackageVersion(InputInterface $in
         return $input->hasOption(self::OPTION_FORCE) && $input->getOption(self::OPTION_FORCE);
     }
 
+    public static function autoInstallBuildTools(InputInterface $input): bool
+    {
+        return $input->hasOption(self::OPTION_AUTO_INSTALL_BUILD_TOOLS)
+            && $input->getOption(self::OPTION_AUTO_INSTALL_BUILD_TOOLS);
+    }
+
+    public static function shouldCheckForBuildTools(InputInterface $input): bool
```
asgrim, line 252: Don't run on Windows...

src/SelfManage/BuildTools/PackageManager.php
```diff
@@ -0,0 +1,103 @@
...
+     * @param list<string> $packages
+     *
+     * @return list<string>
+     */
+    public function installCommand(array $packages): array
+    {
+        return match ($this) {
+            self::Test => ['echo', '"fake installing ' . implode(', ', $packages) . '"'],
+            self::Apt => ['apt-get', 'install', '-y', '--no-install-recommends', '--no-install-suggests', ...$packages],
+            self::Apk => ['apk', 'add', '--no-cache', ...$packages],
```
TimWolla: For Alpine it might be cleaner to pull in all tools by means of a virtual package:

```
apk add --no-cache --virtual .php-pie-deps make bison …
```

asgrim: Thanks, updated :+1:

PR #463, opened 2025-12-19
462: fix intermittent failure to look up packages

### LINKED ISSUE #462: Occasional 'module not found' breaks build, but passes on retry
It seems like for some reason modules can not get found. The one time its the one module, the other time another module. 

Thereafter retrying the build then results in a successful run again.

The output on the reason why the module was not found is not shown, is there a more verbose output mode so I can provide more additional context?

## Missing imagick
Image

## Missing redis
Image

asgrim:
Fixes #462

### DIFF
--- a/src/Installing/InstallForPhpProject/FindMatchingPackages.php
+++ b/src/Installing/InstallForPhpProject/FindMatchingPackages.php
@@ -5,7 +5,7 @@ declare(strict_types=1);
 namespace Php\Pie\Installing\InstallForPhpProject;
 
 use Composer\Composer;
-use Composer\Package\CompletePackage;
+use Composer\Package\CompletePackageInterface;
 use Composer\Repository\RepositoryInterface;
 use OutOfRangeException;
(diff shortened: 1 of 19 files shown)


src/Installing/InstallForPhpProject/FindMatchingPackages.php
```diff
@@ -41,7 +41,7 @@ public function for(Composer $pieComposer, string $searchTerm): array
                 static function (array $match) use ($pieComposer, $extensionName): bool {
                     $package = $pieComposer->getRepositoryManager()->findPackage($match['name'], '*');
 
-                    if (! $package instanceof CompletePackage) {
+                    if (! $package instanceof CompletePackageInterface) {
```
asgrim, line 44: This is the root cause; sometimes (for whatever reason), Composer will return a `CompleteAliasPackage` instead of a `CompletePackage`, which we were discarding. The correct check is now in place, and I can no longer reproduce the reported issue 👍

Issue #436, opened 2025-11-26
Support for prebuilt binaries for extensions


asgrim:
### Progress

 - [x] PIE binary builder GH action - https://github.com/php/pie-ext-binary-builder/
 - [x] PIE 1.3.x forwards-compat PR - php/pie#488 
 - [x] Composer PR (needs 1.3.7 release **first**) - composer/composer#12727
 - [x] Update asgrim/example-pie-extension to build binaries - asgrim/example-pie-extension#12
 - [x] PIE implementation for 1.4.0 - php/pie#489

### Discussed in https://github.com/php/pie/discussions/295

<div type='discussions-op-text'>

<sup>Originally posted by **godruoyi** March 18, 2025</sup>
## Description

Thank you for creating such a nice tool that allows us to avoid using PECL any more. 

I would like to ask if there are any plans to support the direct installation of extension files like .dylib, .dll, and .so, without the need for building them when we run `pie install` . It would be simpler since we can build the extension in another language and publish the artifacts to GitHub, so that we can install them directly without having to build them again. I would be interested in supporting this feature if it is feasible.</div>

### The plan

The proposal is that extensions can provide a matrix pre-built binaries, attached to immutable GH releases, that cover the "80% use case" - i.e. most common combinations. If the combination that is trying to be installed is not available pre-built, then the fallback mechanism of downloading the source & compiling it the traditional way is employed (although the maintainer can disable that behaviour, e.g. if the source code is not open source).

#### My own notes (subject to change! - do NOT quote me on any of this)

 - Ext can specify they provide prebuilt binaries with `DownloadUrlMethod::PrePackagedBinaryDownload`
 - ~~Ext MUST provide a format to build the expected names~~ the format will be fixed (like Windows binaries already are)
 - If we find the asset:
   - Download it
   - Use gh attestations to verify the downloaded asset
      - **note:** depends on implementing ThePHPF/attestation#15
   - Install it (skip build)
   - What to do with metadata? We can generate some (i.e. install metadata), but not the build metadata. Check for reliance on the build metadata
 -  If we do NOT find the asset:
   - Fall back to compiling from source (should this be configurable? e.g. `DownloadUrlMethod::PrePackagedBinaryDownloadWithoutSourceFallback`)
 - Could probably help to create a GH action that builds an ext and attaches prebuilt binaries, to help adoption

#### Approximate matrix of important fields to differentiate builds

 - OS (Windows/Linux/OSX)
[...]

PR #488, opened 2026-01-26
436: add forward compatibility for download-url-method having a list

asgrim:
Since we're planning to allow `download-url-method` to be a list (see #436), this change will allow forward-compatibility with that (otherwise a package that starts to use an array, will crash). This PR pre-empts that (although ideally people should upgrade to 1.4.0 anyway :grin:)


jiggiyDon (commented):
> - **``**gf~~~~

PR #489, opened 2026-01-26
436: pre-packaged binary exts

### LINKED ISSUE #436: Support for prebuilt binaries for extensions

asgrim:
Fixes #436


src/DependencyResolver/Package.php
```diff
@@ -89,7 +92,15 @@ public static function fromComposerCompletePackage(CompletePackageInterface $com
         $package->priority = $phpExtOptions['priority'] ?? 80;
 
         if ($phpExtOptions !== null && array_key_exists('download-url-method', $phpExtOptions)) {
-            $package->downloadUrlMethod = DownloadUrlMethod::tryFrom($phpExtOptions['download-url-method']);
+            /** @var string|list<string> $extOptionValue */
+            $extOptionValue = $phpExtOptions['download-url-method'];
+            $methods        = is_array($extOptionValue) ? $extOptionValue : [$extOptionValue];
```
asgrim, line 97: Needs Composer to be updated & released with this change

asgrim: needs composer/composer#12727 to be merged and released, then we can update Composer dependency to this

asgrim: 2.9.5 was released and this has been updated

PR #509, opened 2026-02-12
Prefix INSTALL_ROOT env var for UnixInstall

### LINKED ISSUE #481: Install extension in specific folder using `INSTALL_ROOT=`
## Bug

Currently it's not possible to install an extension in specific folder by defining `INSTALL_ROOT=`, e.g.:
```shell
INSTALL_ROOT="/your/buildDir" pie install -vj "$(nproc)" apcu/apcu
```

After that you get following:
```shell
In UnixInstall.php line 70:
                                                                                                      
  [RuntimeException]                                                                                  
  Install failed, /usr/local/lib/php/extensions/no-debug-non-zts-20250925/apcu.so was not installed.

Exception trace:
  at phar:///usr/bin/pie/src/Installing/UnixInstall.php:70
```

See following line of code:
https://github.com/php/pie/blob/1.4.x/src/Installing/UnixInstall.php#L69

## Reason

The variable `INSTALL_ROOT` is already working fine with the extensions (while build-process), just pie throws the exception above because it does not handle any installation folders.
In my opinion there should be a pie command option `--install-dir|--prefix` (prefix should then also be used for the php binary paths) defining an installation directory and bypassing this value to the sys env `INSTALL_ROOT` before running the builds.

## Workaround

My current workaround is building twice:

```
pie install -vj "$(nproc)" apcu/apcu
INSTALL_ROOT="/your/buildDir" pie install -vj "$(nproc)" apcu/apcu
```

Works like a charm, but yeah, I'm building twice...

asgrim:
Fixes #481

### DIFF
--- a/src/Installing/UnixInstall.php
+++ b/src/Installing/UnixInstall.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace Php\Pie\Installing;
 
 use Composer\IO\IOInterface;
+use Composer\Util\Platform as ComposerPlatform;
 use Php\Pie\Downloading\DownloadedPackage;
 use Php\Pie\Downloading\DownloadUrlMethod;
 use Php\Pie\File\BinaryFile;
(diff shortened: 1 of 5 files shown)


asgrim:
~~Arm runners seem to be dead ATM, see actions/partner-runner-images#166~~ seems fixed now 25/2

PR #512, opened 2026-02-19
workaround zlib 1.3.2 library naming change on Windows

### LINKED ISSUE #514: Windows CI binary build fails due to zlib 1.3.2 library naming change
The `build-executable` workflow fails on Windows because zlib 1.3.2 changed its cmake output file names:

| Old (<=1.3.1) | New (1.3.2) |
|---|---|
| `zlibstatic.lib` | `zs.lib` |
| `zlib.lib` | `z.lib` |
| `zlib.dll` | `z.dll` |

SPC's zlib builder expects the old names, so the static library never gets installed correctly. PHP's configure can't find zlib, and the final link fails with 12 unresolved symbols (inflate, deflate, etc.) from curl, openssl, and libssh2.

```
libcurl_a.lib(content_encoding.obj) : error LNK2019: unresolved external symbol inflate
libcurl_a.lib(content_encoding.obj) : error LNK2019: unresolved external symbol deflateEnd
...
```

Upstream issue filed: https://github.com/crazywhalecc/static-php-cli/issues/1039

In the meantime the workflow needs a workaround to rename the files and re-run configure + nmake.

GhostPirateBob:
Fixes #514

This is a bit messy but see the issue for the rationale. I'm hoping this is just a temporary workaround and you can get rid of it after other fixes upstream.

Upstream issue filed: https://github.com/crazywhalecc/static-php-cli/issues/1039

zlib 1.3.2 changed its cmake output file names on Windows:
- zlibstatic.lib -> zs.lib
- zlib.lib -> z.lib
- zlib.dll -> z.dll

SPC's zlib builder expects the old names, and PHP's configure looks for zlib_a.lib or zlib.lib. With neither present, zlib is silently excluded and the link step fails with 12 unresolved zlib symbols from curl, openssl, and libssh2.

This adds a conditional workaround that renames the files and re-runs configure + nmake when the issue is detected. It can be removed once static-php-cli updates their zlib builder.

Cheers!

gpb
✌(◕‿-)✌

### DIFF
--- a/.github/workflows/build-assets.yml
+++ b/.github/workflows/build-assets.yml
@@ -129,9 +129,50 @@ jobs:
           name: pie-${{ github.sha }}.phar
 
       - name: Build for ${{ runner.os }} ${{ runner.arch }} on ${{ matrix.operating-system }}
+        if: runner.os != 'Windows'
         run: ${{ env.SPC_BINARY }} craft resources/spc/craft.yml
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Build for ${{ runner.os }} ${{ runner.arch }} on ${{ matrix.operating-system }}
+        if: runner.os == 'Windows'
+        shell: pwsh
+        run: |
+          # Run craft — may fail at link time due to zlib 1.3.2 renaming its output files.
+          # See workaround below. Exit 0 so the workflow continues either way.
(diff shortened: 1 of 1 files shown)


asgrim:
Sorry, but I'd rather not have such a complicated workaround added here; I have disabled the Windows build for now in #516 until the upstream issue can be addressed. Thanks anyway!

PR #515, opened 2026-02-19
Fix Windows extension install from downloads.php.net

### LINKED ISSUE #513: Windows extension install fails due to architecture naming mismatch (x86_64 vs x64)
Installing extensions on Windows can fail when the release asset uses `x64` in the filename instead of `x86_64`, or omits the `v` prefix from the version string.

For example, `pie install apcu/apcu` on Windows fails because PIE looks for `php_apcu-v5.1.28-8.5-ts-vs17-x86_64.zip` but the actual asset on downloads.php.net is named `php_apcu-5.1.28-8.5-ts-vs17-x64.zip` (different arch label, no `v` prefix).

Similarly, the DLL inside the zip is just `php_apcu.dll` (no version/platform suffix), which PIE doesn't check for.

```
> php pie-1.4.x-nightly.phar install apcu/apcu

🥧 PHP Installer for Extensions (PIE) 1.3.8@adfe984, from The PHP Foundation
This command may need elevated privileges, and may prompt you for your password.
You are running PHP 8.5.1
Target PHP installation: 8.5.1 ts, vs17, on Windows x86_64 (from C:\FlyEnv\PhpWebStudy-Data\env\php\php.exe)
Found package: apcu/apcu:v5.1.28 which provides ext-apcu

In CouldNotDetermineDownloadUrlMethod.php line 31:

  Could not download apcu/apcu using windows-binary method: Windows archive with prebuilt extension for apcu/apcu was not attached on release v5.1.28 - looked for one of "php_apcu-v5.1.28-8.5-ts-vs17-x86_64.zip, php_apcu-v5.1.28-8.5-vs17-ts-x86_64.zip"
```

**Environment:** PIE 1.3.8 / nightly, Windows, PHP 8.5.1 ts vs17 x86_64

GhostPirateBob:
Fixes #513

G'day folks!

This is my first contribution to this repo, so please let me know if there’s a preferred way to structure things.

I had a weird issue on Windows with the latest nightly and 1.3.8 stable where `pie install apcu/apcu` didn't work, even though there was a valid zip file and dll for my version.

Anyway turns out it was to do with inconsistent naming across different $arch so I made a PR to fix it.

I also ran into another weird build workflow issue while doing this PR, but I split it off into a second one because it was somewhat unrelated (see #514 and #512).

Cheers!

gpb
✌(◕‿-)✌

## Summary

- Adds `Architecture::allNames()` returning all known name variants per architecture (e.g. `x86_64` + `x64`, `arm64` + `aarch64`), so asset filename matching works across platforms that use different conventions
- Updates `WindowsExtensionAssetName` to generate filename variants covering all arch aliases and version formats (with/without `v` prefix), fixing installs from sources like downloads.php.net
- Updates `PrePackagedBinaryAssetName` and `GithubPackageReleaseAssets` to use `allNames()` for broader asset matching
- Adds fallback for simple DLL names (`php_<ext>.dll`) used by downloads.php.net

After fix:

### DIFF
--- a/src/Downloading/GithubPackageReleaseAssets.php
+++ b/src/Downloading/GithubPackageReleaseAssets.php
@@ -12,6 +12,7 @@ use Webmozart\Assert\Assert;
 
 use function array_map;
 use function in_array;
+use function ltrim;
 use function strtolower;
 
 /** @internal This is not public API for PIE, so should not be depended upon unless you accept the risk of BC breaks */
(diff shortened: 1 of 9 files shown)


asgrim:
As per #513 - closing this for now;

> The correct fix should be that `apcu` should add the Windows builds to the releases, as per the PIE docs: https://github.com/php/pie/blob/1.4.x/docs/extension-maintainers.md#windows-support
> 
> I don't think PIE should be updated to fall back to downloading from `downloads.php.net` (like [#515](https://github.com/php/pie/pull/515)) as (like with PECL itself) we don't really want to be maintaining this longer term, but I'll see what the team thinks, so it is unlikely we would merge that PR.

PR #518, opened 2026-02-25
Add new PHP extensions to supported list

macintoshplus:
Added new PHP extensions 'rar', 'sync', and 'varnish' to the supported extensions list.


asgrim (approved):
Thanks!

PR #545, opened 2026-03-16
Allow distribution of musl + glibc binaries

### LINKED ISSUE #544: Search for Linux binaries without libc suffix
### Discussed in https://github.com/php/pie/discussions/543

<sup>Originally posted by **cataphract** March 16, 2026</sup>
The libc suffix in the binary names forces one to have separate musl and glibc binaries, or at least to have the same binary available with two names.

What do you think of looking also for a binary without the libc suffix so we can more efficiently ship binaries that work for both?</div>

cataphract:
Search for binaries with the `anylibc` suffix as a fallback candidate (on Linux only).

This allows shipping binaries that work both on glibc and musl without copying the same binary with two separate names.

The second commit fixes assertion failures when not running a linux/glibc (I think a better long-term solution would be to extend `TargetPlatform` to also take the libc flavor.

Fixes #544 

## PR submitter checklist

- [x] I have read [CONTRIBUTING.md](../CONTRIBUTING.md)
- [x] I discussed this feature with the maintainers in #544

### DIFF
--- a/src/Platform/PrePackagedBinaryAssetName.php
+++ b/src/Platform/PrePackagedBinaryAssetName.php
@@ -21,10 +21,10 @@ final class PrePackagedBinaryAssetName
     /** @return non-empty-list<non-empty-string> */
     public static function packageNames(TargetPlatform $targetPlatform, Package $package): array
     {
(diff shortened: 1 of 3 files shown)


asgrim (approved):

asgrim:
Thanks @cataphract :)

PR #553, opened 2026-03-18
Add mutation testing with Infection

asgrim:
## PR submitter checklist

- [x] I have read [CONTRIBUTING.md](../CONTRIBUTING.md)
- [x] I discussed this feature with the maintainers in my own head

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -203,6 +203,22 @@ jobs:
       - name: Run phpstan
         run: vendor/bin/phpstan
 
+  mutation-testing:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Setup PHP
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: 8.4
+          extensions: intl, sodium, zip
+          tools: infection
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      - uses: actions/checkout@v6
+      - uses: ramsey/composer-install@v3
+      - name: Run infection
+        run: sudo infection --min-msi=68 --min-covered-msi=68 --threads=4 --show-mutations=max --no-progress
+
   build-assets:
     needs:
       - unit-tests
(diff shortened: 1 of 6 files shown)


.github/workflows/continuous-integration.yml
```diff
@@ -203,6 +203,22 @@ jobs:
...
+        with:
+          php-version: 8.4
+          extensions: intl, sodium, zip
+          tools: infection
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      - uses: actions/checkout@v6
+      - uses: ramsey/composer-install@v3
+      - name: Run infection
+        run: sudo infection --min-msi=68 --min-covered-msi=68 --threads=4 --show-mutations=max --no-progress
```
theofidry: - is `sudo` necessary here?!
- any reason to use `--threads=4` instead of `--threads=max`?

infection.json5
```diff
@@ -0,0 +1,11 @@
+{
+    "$schema": "https://raw.githubusercontent.com/infection/infection/0.32.6/resources/schema.json",
+    "source": {
+        "directories": [
+            "src"
+        ]
+    },
+    "mutators": {
+        "@default": true
```
theofidry, line 9: as an app, maybe it's already disabled by default, there is the public -> protected/private mutator that can be disabled

infection.json5
```diff
@@ -0,0 +1,11 @@
+{
+    "$schema": "https://raw.githubusercontent.com/infection/infection/0.32.6/resources/schema.json",
```
theofidry, line 2: so what's the workflow to run it locally and not the CI when you're developing?

.github/workflows/continuous-integration.yml
asgrim: >* is `sudo` necessary here?!

yep, the tests require it (because I can't non-interactively `sudo`)

>* any reason to use `--threads=4` instead of `--threads=max`?

nope :) will update

infection.json5
asgrim: use the phar, as per the docs; I suppose that should be documented :grin:

infection.json5
asgrim: Does `@internal` make that ignored anyway? Hopefully all of the classes have `@internal` annotations anyway

PR #558, opened 2026-03-19
add installation instructions for RPM packages

remicollet:
- add a `Distribution packages` section in the usage page
- add RPM case 

Other distro may be added later by their maintainers

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -31,6 +31,19 @@ curl -fL --output /tmp/pie.phar https://github.com/php/pie/releases/latest/downl
   && sudo chmod +x /usr/local/bin/pie
 ```
 
+### Distribution packages
+
+**On Fedora and Enterprise Linux 10**
+
+On Enterprise Linux (CentOS, RHEL, AlmaLinux, RockyLinux, and other clones) you
+need to enable the [EPEL](https://docs.fedoraproject.org/en-US/epel/) repository.
+
+```shell
+sudo dnf install pie
+```
+
+Package information: [pie](https://src.fedoraproject.org/rpms/pie)
+
 ### Docker installation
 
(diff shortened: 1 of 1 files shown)


remicollet:
Amended for minor fix: Only EL-10 is supported

(because EL-8 has PHP 7.4 and EL-9 has PHP 8.0 by default)

asgrim (approved):

docs/usage.md
```diff
@@ -33,17 +33,35 @@ curl -fL --output /tmp/pie.phar https://github.com/php/pie/releases/latest/downl
...
+#### Fedora and Enterprise Linux 10
 
 On Enterprise Linux (CentOS, RHEL, AlmaLinux, RockyLinux, and other clones) you
 need to enable the [EPEL](https://docs.fedoraproject.org/en-US/epel/) repository.
 
 ```shell
 sudo dnf install pie
+
+# If you have `gh`, you can verify PIE is authentic:
+gh attestation verify --owner=php $(which pie)
```
remicollet, line 49: This won't work for the RPM package, which is not the phar archive
It is signed by the build provider and can be verified using the` rpm --verify` command (always checked on install/update)

asgrim: This is not verifying the RPM, it is verifying the PHAR; `which pie` should return `/usr/bin/pie` in this case, which should be verifiable against the original owner (unless you have built the PHAR yourself, which I would not endorse)

remicollet: `/usr/bin/pie` is NOT the phar, so cannot be verified by the gh attestations

The package contains the "composer install" result tree (`/usr/share/pie`)
And the pie command is the one from this tree.

asgrim: Thank you - I made a PR to remove this in #574

Issue #480, opened 2026-01-08
Offer fine-grained (major/minor specific) Docker image tags


jnoordsij:
Currently the Docker images published by this package come in two flavors:
- the latest version (called `bin`)
- version-specific builds (e.g. `1.3.5-bin` or `1.3.0-rc.3` for previews)

To allow developers to have more flexibility in in what version to employ, it would be nice to have major and minor-specific tags as well, so that one has more choice in the desired 'stability' of the consumed image.

asgrim:
ref: https://github.com/php/pie/pull/614#issuecomment-4544122792

> This fixed #480, right?
> 
> However, I wonder if this implementation is entirely correct: for e.g. 1.5.0-alpha.1, this would now also update the `1-bin` 'major' image if I understand things correctly.
> 
> Likely you'd want to distinguish stable and non-stable tags in some way, similar to how it's done for the floating `bin` tag.

PR #614, opened 2026-05-13
Add granular publish tags for "-bin" docker image

glensc:
Implement TODO:
- https://github.com/php/pie/blob/5c3fa89c5f63728f60fda04756b4044c01855ff7/.github/workflows/release.yml#L90-L96

See:
- https://github.com/php/pie/pull/122#pullrequestreview-2477496308
- https://github.com/php/pie/pull/122#discussion_r1867331273
- https://github.com/php/pie/issues/480

## PR submitter checklist

- [x] I discussed this feature with the maintainers in https://github.com/php/pie/pull/122
- [x] <del>I have added appropriate tests</del>

## Future TODO

After this is merged and verified working also update documentation (another PR):
- https://github.com/php/pie/blob/1.5.x/docs/usage.md#docker-installation

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -88,9 +88,8 @@ jobs:
       packages: write
     uses: ./.github/workflows/build-and-push-docker-image.yml
     with:
-      # @TODO v1.0 Consider introducing more granular tags (major and major.minor)
-      # @see https://github.com/php/pie/pull/122#pullrequestreview-2477496308
-      # @see https://github.com/php/pie/pull/122#discussion_r1867331273
       tags: |
         ${{ ((!contains(github.ref, 'alpha') && !contains(github.ref, 'beta') && !contains(github.ref, 'rc')) && 'type=raw,value=bin') || '' }}
         type=semver,pattern={{version}},suffix=-bin
+        type=semver,pattern={{major}}.{{minor}},suffix=bin
+        type=semver,pattern={{major}},suffix=bin
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -91,5 +91,5 @@ jobs:
       tags: |
         ${{ ((!contains(github.ref, 'alpha') && !contains(github.ref, 'beta') && !contains(github.ref, 'rc')) && 'type=raw,value=bin') || '' }}
         type=semver,pattern={{version}},suffix=-bin
-        type=semver,pattern={{major}}.{{minor}},suffix=bin
-        type=semver,pattern={{major}},suffix=bin
+        type=semver,pattern={{major}}.{{minor}},suffix=-bin
+        type=semver,pattern={{major}},suffix=-bin


asgrim:
@glensc please can you complete the PR submitter checklist; you can reference #122 in the second checklist item

glensc:
@asgrim ✅

.github/workflows/release.yml
```diff
@@ -88,9 +88,8 @@ jobs:
       packages: write
     uses: ./.github/workflows/build-and-push-docker-image.yml
     with:
-      # @TODO v1.0 Consider introducing more granular tags (major and major.minor)
-      # @see https://github.com/php/pie/pull/122#pullrequestreview-2477496308
-      # @see https://github.com/php/pie/pull/122#discussion_r1867331273
       tags: |
         ${{ ((!contains(github.ref, 'alpha') && !contains(github.ref, 'beta') && !contains(github.ref, 'rc')) && 'type=raw,value=bin') || '' }}
         type=semver,pattern={{version}},suffix=-bin
+        type=semver,pattern={{major}}.{{minor}},suffix=bin
```
asgrim: I haven't checked this, but i think the `suffix` for both of these should be `-bin` (see the line above); otherwise I think the tags would be `2.1.5-bin`, `2.1bin`, `2bin` for example, it would be better/more consistent to use the hypen IMO

jnoordsij:
This fixed #480, right?

However, I wonder if this implementation is entirely correct: for e.g. 1.5.0-alpha.1, this would now also update the `1-bin` 'major' image if I understand things correctly. 

Likely you'd want to distinguish stable and non-stable tags in some way, similar to how it's done for the floating `bin` tag.

asgrim:
> This fixed #480, right?
> 
> However, I wonder if this implementation is entirely correct: for e.g. 1.5.0-alpha.1, this would now also update the `1-bin` 'major' image if I understand things correctly.
> 
> Likely you'd want to distinguish stable and non-stable tags in some way, similar to how it's done for the floating `bin` tag.

yep

.github/workflows/release.yml
glensc: Yeah, makes sense. I din't test (too difficult to setup test), I just copied them from elsewhere.

glensc:
> This fixed #480, right?
> 
> However, I wonder if this implementation is entirely correct: for e.g. 1.5.0-alpha.1, this would now also update the `1-bin` 'major' image if I understand things correctly.
> 
> Likely you'd want to distinguish stable and non-stable tags in some way, similar to how it's done for the floating `bin` tag.

I suppose similar guard can added there:

```yml
${{ ((!contains(github.ref, 'alpha') && !contains(github.ref, 'beta') && !contains(github.ref, 'rc')) && 'type=raw,value=bin') || '' }}
```

glensc:
This also needs to be backported to 1.4.x branch?

TimWolla:
Besides the obvious bug with pre-release versions, I wonder what problem this is attempting to solve. Is there ever a good reason to intentionally stay on an outdated branch of PIE?

glensc:
Well, this is merged to 1.5, but there are no 1.5 releases.

I'm new to pie, so i'm not aware when will be 1.5.0 is released, and i would like to update my Dockerfile to 1.x branch, be it 1.4 or 1.5, not really relevant to me. i'm assuming the project follows basic semver.

.github/workflows/release.yml
glensc: Seems it was fixed already on the same branch: https://github.com/php/pie/commit/afa872506d23a850ab3042051cefa537e47c4ddf

glensc:
> > This fixed #480, right?
> > However, I wonder if this implementation is entirely correct: for e.g. 1.5.0-alpha.1, this would now also update the `1-bin` 'major' image if I understand things correctly.
> > Likely you'd want to distinguish stable and non-stable tags in some way, similar to how it's done for the floating `bin` tag.
> 
> I suppose similar guard can added there:
> 
> ```yaml
> ${{ ((!contains(github.ref, 'alpha') && !contains(github.ref, 'beta') && !contains(github.ref, 'rc')) && 'type=raw,value=bin') || '' }}
> ```

Fix:
- https://github.com/php/pie/pull/620

It maybe not loking the nicest, it follows the same pattern, so should be easy to review

TimWolla:
> and i would like to update my Dockerfile to 1.x branch,

Yes. I am asking why we need a `1.4-bin` tag. This will just lead to folks staying on outdated branches of PIE.

glensc:
Since there is no 1.5-bin (no 1.5.x stable releases), i'd use what's available, as 1.4 seems active, so 1.4-bin

Correction:
- I'd use `1-bin`, but that would require backport to 1.4 (and a release) or 1.5.x stable release

PR #617, opened 2026-05-15
Various Hardening Updates

iliaal:
## Summary

Five hardening changes from a security audit. None is a fresh attacker primitive on its own; the audit recommended shipping them as one PR.

- `RemoveIniEntry`: anchor the regex with `\b` and `preg_quote` the extension name, so uninstalling `foo` no longer rewrites the prefix of `extension=foo_other` lines.
- `PlaceholderReplacer` and `WindowsInstall`: skip symlinks during the recursive source walk, so the iterator no longer descends into files outside the extracted source dir.
- `OverrideDownloadUrlInstallListener`: warn that swapping to a release-asset URL leaves only HTTPS-to-origin as the integrity guarantee, since Composer's dist-sha was bound to the original Packagist URL.
- `ConfigureOption`: validate `php-ext.configure-options[].name` against `/^[a-zA-Z][a-zA-Z0-9_-]*$/`, so configure-flag identifiers can't carry whitespace or shell metacharacters into argv or `installed.json`.
- `FallbackVerificationUsingOpenSsl`: promote the fallback notice to `writeError` and document that the OpenSSL path skips Rekor transparency-log verification.

### DIFF
--- a/src/Building/PlaceholderReplacer.php
+++ b/src/Building/PlaceholderReplacer.php
@@ -57,6 +57,13 @@ class PlaceholderReplacer
         foreach (new RecursiveIteratorIterator(new RecursiveDirectoryIterator($downloadedPackage->extractedSourcePath)) as $file) {
             assert($file instanceof SplFileInfo);
 
+            // Refuse to follow symlinks into files the package author did not
+            // legitimately own at build time (e.g., an archive entry that points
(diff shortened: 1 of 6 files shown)


asgrim (approved):
Thanks! 👍

PR #651, opened 2026-06-29
refactor: enhance UX for missing build tools prompt

yilanboy:
This is a little UX improvement PR. When a tool is missing, pie will ask the user for installation conformation, but it lacks some instruction to tell the user what to do next and what is the default action.

So I just update the question prompt from:

```text
Would you like to install them now?
````

to:

```text
Would you like to install them now? [y/N]
```

The `[y/N]` tells the user can type `y` or `n`, and `n` is the default action.

## PR submitter checklist

- [x] I discussed this feature with the maintainers in #649 
- [x] I confirm that I have the right to submit this under the project's open source license


asgrim (approved):
Thank you, spot on! 👍

PR #652, opened 2026-06-30
587: add `pie install --from-lock`

### LINKED ISSUE #587: Add pinning of installed packages and documentation regarding pie.lock file
### Discussed in https://github.com/php/pie/discussions/575

<sup>Originally posted by **cleboo** April  2, 2026</sup>
Hi!

I wanted suggest an improvement of the docs of pie when it comes to pinning of dependencies.

The tl;dr is this:
+ Add documentation on pie.lock files and the behaviour of `pie install`. Does it check against the lock-file? Or does it always get the newest version? What do I have to do to make sure that I build from the the exact same source code I installed from the last time, pinning some kind of checksum?

I am looking for a way to pin the code/artifacts that are downloaded and installed by pie to make sure, we don't unintentionally install new code. As far as I understand, right now pie allows us to
+ pin to versions or version ranges - but that does not protect against the code being replaced underneath the tag (pinned tags on github+attestations may help here, but that has to be configured for every extension separately).
+ install a specific git commit with `pie install xdebug/xdebug:dev-master#af9280fff4eeb014309f8f66633c5e8f40e88b04` (this makes upgrading a hassle).

I noticed, that there is a pie.lock file generated in the `$HOME/.pie/php<version>_<some hash>` and that it contains a `dist.shasum` field (that is currently empty), as well as the exact version installed, but I could not find any documentation on whether this actually pins installed source code or not. When I run `pie install`, will it try to upgrade to newer code or install the code that is mentioned in pie.lock?

I guess that the idea of pinning the installed code is not new and that implementing an `install/update` behavior similar to composer is planned, since the lock file is already there. But adding some documentation around the `pie install` command for now that makes clear what it does when a pie.lock file is already present, would be good to avoid confusion.

Thanks for your time!

For reference: the discussion on [verifying attestations during install](https://github.com/php/pie/discussions/567) is also an important aspect of this.
</div>

asgrim:
Fixes #587 

## PR submitter checklist

- [x] I discussed this feature with the maintainers in #587 

### DIFF
--- a/features/install-extensions.feature
+++ b/features/install-extensions.feature
@@ -14,3 +14,9 @@ Feature: Extensions can be installed with PIE
   Example: Multiple extensions can be installed at once
     When I run a command to install multiple extensions
     Then the extensions should have been installed and enabled
+
+  # pie install --from-lock
+  Example: I can install exact versions contained in the lockfile
+    Given I have a lock file
+    When I run a command to install from the lockfile
+    Then the extensions should have been updated to the lock
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -160,6 +160,7 @@ jobs:
           - '8.2'
           - '8.3'
           - '8.4'
+          - '8.5'
     steps:
       - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 #v6.0.3
         with:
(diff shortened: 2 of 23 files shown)


features/install-extensions.feature
```diff
@@ -14,3 +14,9 @@ Feature: Extensions can be installed with PIE
   Example: Multiple extensions can be installed at once
     When I run a command to install multiple extensions
     Then the extensions should have been installed and enabled
+
+  # pie install --from-lock
```
asgrim, line 18: Need to review/update docs

PR #667, opened 2026-07-10
658: PHP 8.6 (pre-release) support

### LINKED ISSUE #658: Add support for PHP 8.6 (pre-release)
Given the first alpha has landed, starting to test with PHP 8.6 might be appropriate. Although formal support may be delayed until 8.6 has stabilized further (e.g. beta or RC), at least tracking compatibility on a PR is probably useful.

asgrim:
Fixes #658 

## PR submitter checklist

- [x] I discussed this feature with the maintainers in #658 


.github/workflows/continuous-integration.yml
```diff
@@ -21,6 +21,7 @@ jobs:
           - '8.3'
           - '8.4'
           - '8.5'
+          - '8.6'
```
asgrim, line 24: needs test ext to be released with 8.6, as Windows builds don't exist for 8.6 yet

see asgrim/example-pie-extension#13

failing tests - https://github.com/php/pie/actions/runs/29113092108/job/86429844134#step:7:32

.github/workflows/continuous-integration.yml
```diff
@@ -62,6 +63,7 @@ jobs:
           - '8.3'
           - '8.4'
           - '8.5'
+          #- '8.6' # @todo bundled repo doesn't support non-stable PHP versions
```
asgrim, line 125: tracking todo

.github/workflows/continuous-integration.yml
```diff
@@ -161,6 +170,7 @@ jobs:
           - '8.3'
           - '8.4'
           - '8.5'
+          - '8.6-dev'
```
asgrim: failing because the test image deps on xdebug/xdebug which doesn't exist for 8.6 (yet)

https://github.com/php/pie/actions/runs/29113092108/job/86429844062#step:6:2673

that said, in theory we could remove that requirement, but it is useful for local debugging 🤔

jnoordsij: FWIW I had similar issues in the past on my testing images when trying to install this, but with PIE I can now do `pie install "xdebug/xdebug:dev-master"`. Although not a formal release, for testing purposes on pre-release images I considered this fine. Otherwise you'd have to wait for a first (pre-)release of XDebug compatible with PHP 8.6, which likely will take a while to land.

asgrim: Hm; yeah I did try this, but I realised, for 8.5 I had disabled the behaviour tests because there's a few other changes that would be needed. Think I'll just not run them for 8.6 for now, and come back to it once stable or have more time :D

PR #671, opened 2026-07-13
650: support gmake on freebsd

### LINKED ISSUE #650: On FreeBSD, gmake instead of make should be used
I have tried running

`pie install mongodb/mongodb-extension`

on FreeBSD what resulted to following error:

>  make: /root/.pie/php8.5_e5f4aa0c439f2f7b3b779be755bdf64c/vendor/mongodb/mongodb-extension/mongodb-2.3.3/Makefile:208: warning: Invalid character " " in variable name "realpath /usr/local/lib/php/20250925/json.so"

On FreeBSD, simple **make** command runs BSD make, which is available as a part of base system on **/usr/bin/make** path.

GNU make has to be installed as a package, its installation path is **/usr/local/bin/gmake** and could be run as a **gmake** command.

As a test and workaround, I made a temporary symlink **/sbin/make -> /usr/local/bin/gmake** which caused that executing **make** runs GNU make, as expected by pie. After this change I was able to run 

`pie install mongodb/mongodb-extension`

without any issue.

So, to solve this bug, pie should do some test to detect its environment and choose between **make** and **gmake** correctly. It could be running **uname** to detect FreeBSD as an OS, it could be running **make -v** and **gmake -v** and check for "GNU make" string, or something similar.

Or, at least, command line option called **--with-make-path** should be added, like there are similar options for phpize, php-config and php.

Maybe it would be good to do both.

asgrim:
Fixes #650

## PR submitter checklist

- [x] I discussed this feature with the maintainers in #650 


src/Platform/MakePath.php
```diff
@@ -0,0 +1,66 @@
...
+     * Make syntax. On some platforms (e.g. FreeBSD, NetBSD, OpenBSD), the
+     * `make` found on the PATH is the native BSD make, which cannot parse
+     * these Makefiles; GNU Make is available there as `gmake` instead.
+     *
+     * @return non-empty-string
+     */
+    public static function guess(): string
+    {
+        foreach (['make', 'gmake'] as $candidate) {
+            $which = new Process(['which', $candidate]);
```
theofidry, line 48: IIRC it could be that it inherits the current process verbosity, in which case if used in quiet mode for example you may not have an output there when you would expect to.

I vaguely remember that problem in Box and I'm about to fix some instances I noticed in Infection, but it's not clear 100% to me when it occurs so you may not be hit by this case.

asgrim: @theofidry is that specific to FreeBSD? or just in general? I've not observed this verbosity inheritance behaviour before... :thinking:

theofidry: Ok I verified it because I feel like an ass just throwing a random claim. You can see it [here](https://github.com/symfony/symfony/blob/8.2/src/Symfony/Component/Process/Process.php#L1713):

```php
    private function getDefaultEnv(): array
    {
        $env = getenv();
        $env = ('\\' === \DIRECTORY_SEPARATOR ? array_intersect_ukey($env, $_SERVER, 'strcasecmp') : array_intersect_key($env, $_SERVER)) ?: $env;

        return $_ENV + ('\\' === \DIRECTORY_SEPARATOR ? array_diff_ukey($env, $_ENV, 'strcasecmp') : $env);
    }
```

which is used in `::start()` therefore `::run()` and `::mustRun()` too.

So if `SHELL_VERBOSITY` is set somehow, which can be explicitly by the user or using `--quiet` when using the Symfony Console (the Application sets the environment variable, see [this piece of code](https://github.com/symfony/symfony/blob/8.2/src/Symfony/Component/Console/Application.php#L1035)).

Small reproducer:

```php
<?php // demo.php

require __DIR__.'/vendor/autoload.php';

use Symfony\Component\Process\Process;

// Simulate SHELL_VERBOSITY having been set by the parent application.
putenv('SHELL_VERBOSITY=-1');
$_ENV['SHELL_VERBOSITY'] = '-1';
$_SERVER['SHELL_VERBOSITY'] = '-1';

$command = [
    PHP_BINARY,
    '-r',
    'echo getenv("SHELL_VERBOSITY"), PHP_EOL;',
];

// No environment explicitly provided: SHELL_VERBOSITY is inherited.
$inheritedProcess = new Process($command);
$inheritedProcess->mustRun();

echo 'Inherited: '.$inheritedProcess->getOutput();

// Explicitly reset the verbosity for the child process.
$overriddenProcess = new Process(
    command: $command,
    env: ['SHELL_VERBOSITY' => '0'],
);
$overriddenProcess->mustRun();

echo 'Overridden: '.$overriddenProcess->getOutput();
```

Then:

```
$ php demo.php
Inherited: -1
Overridden: 0
```

asgrim: Thanks, I created #682 to look into this :+1:

PR #704, opened 2026-08-08
Fix build provider detection for Remi

remicollet:
The Remi build provider string ends with a hashtag, but this may change

Ex: https://git.remirepo.net/cgit/tools/mock.git/diff/vendor.tpl?id=4f969f42ccd857ed9366c7b74fc6e587cc4ae56d

### DIFF
--- a/src/DependencyResolver/ResolveDependencyWithComposer.php
+++ b/src/DependencyResolver/ResolveDependencyWithComposer.php
@@ -114,7 +114,7 @@ final class ResolveDependencyWithComposer implements DependencyResolver
             throw BundledPhpExtensionRefusal::forPhpExtraVersion($targetPlatform->phpBinaryPath);
         }
 
-        $buildProvider           = $targetPlatform->phpBinaryPath->buildProvider();
+        $buildProvider = $targetPlatform->phpBinaryPath->buildProvider();
         if (! $buildProvider) {
             return;
         }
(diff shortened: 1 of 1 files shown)


src/Platform/TargetPhp/PhpBinaryPath.php
```diff
@@ -210,9 +212,12 @@ public function loadedIniConfigurationFile(): string|null
     public function buildProvider(): string|null
     {
         /**
-         * Newer versions of PHP will have a `PHP_BUILD_PROVIDER` constant
-         * defined - {@link https://github.com/php/php-src/pull/19157}
+         * PHP 8.5+ have a `PHP_BUILD_PROVIDER` constant defined
          */
+        if (defined('PHP_BUILD_PROVIDER')) {
```
asgrim: This approach won't really work because the `PhpBinaryPath` we're installing the extension to isn't necessarily the one running PIE; an example of using a constant from the actual target PHP install in this manner can be seen in `\Php\Pie\Platform\TargetPhp\PhpBinaryPath::phpVersionWithExtra`

asgrim (changes requested):

src/Platform/TargetPhp/PhpBinaryPath.php
remicollet: OK, last commit removed
BTW, the comment about PHP_BUILD_PROVIDER is confusing ;)

asgrim: Yeah I can see that now, it was only worded/useful for my own memory, which isn't ideal!

asgrim (approved):

asgrim:
Thanks! :shipit:

